PR #27, opened 2024-08-09
Add GHA workflow to run lint and tests

mfiedorowicz:
Adding GitHub Actions workflow to run lint (with `ruff`), building docs (`mkdocs build`) and tests


.github/lint-tests.yml
```diff
@@ -0,0 +1,44 @@
...
+  checks: write
+  pull-requests: write
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        python: [ "3.10" ]
```
mfiedorowicz: Starting with `3.10` for time being, we can expand it with `3.11` and `3.12` once workflow is working properly.

pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+]
+package-data = {"netbox_branching" = ["**/*", "templates/**"]}
+exclude-package-data = {netbox_branching = ["tests/*"]}
+license-files = ["LICENSE.md"]
+
+[build-system]
+requires = ["setuptools>=43.0.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[tool.ruff]
```
jeremystretch: Let's disable the checks for docstrings to cut down on noise. (These sometimes aren't needed, and their formatting is standard.)

.github/workflows/lint-tests.yml
```diff
@@ -0,0 +1,44 @@
...
+  pull-requests: write
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        python: [ "3.10" ]
+    steps:
```
jeremystretch: We should include a step that builds the documentation using `mkdocs build` as well.

pyproject.toml
mfiedorowicz: Disabled ~most~ all checks for `missing docstrings`.

.github/workflows/lint-tests.yml
mfiedorowicz: Where should we build into?

mfiedorowicz: Leaving in the default location for now (`./site`)

jeremystretch:
I'm concerned about the introduction of a Docker container for testing and all the additional complexity it brings. Does this buy us anything that we wouldn't get with a [much simpler workflow](https://github.com/peteeckel/netbox-plugin-dns/blob/main/.github/workflows/test.yml) running NetBox natively?

mfiedorowicz:
> I'm concerned about the introduction of a Docker container for testing and all the additional complexity it brings. Does this buy us anything that we wouldn't get with a [much simpler workflow](https://github.com/peteeckel/netbox-plugin-dns/blob/main/.github/workflows/test.yml) running NetBox natively?

Agreed, especially that we can run tests against various python versions.

jeremystretch (approved):

PR #58, opened 2024-08-20
Fixes #57: Record ChangeDiffs only for supported object types

### LINKED ISSUE #57: ChangeDiff records are created automatically for excluded objects
### Plugin Version

v0.3.1

### NetBox Version

v4.1-beta1

### Python Version

3.10

### Steps to Reproduce

The signal receiver which watches for object events and [creates ChangeDiffs](https://github.com/netboxlabs/netbox-branching/blob/ade37d175fe3fd9bf2244660d85623fc17bdbe35/netbox_branching/signal_receivers.py#L62) in response does not currently filter out object types that should be excluded.

### Expected Behavior

 ChangeDiffs should be created/updated only for supported models, e.g. those returned by `get_branchable_object_types()`.

### Observed Behavior

A ChangeDiff is created/updated even for models which don't support branching.

jeremystretch:
### Fixes: #57

- Move excluded apps to a constant
- Add an entry for `branching` in the model features registry
- Modify `record_change_diff()` to skip unsupported object types

### DIFF
--- a/netbox_branching/__init__.py
+++ b/netbox_branching/__init__.py
@@ -1,4 +1,5 @@
 from netbox.plugins import PluginConfig
+from netbox.registry import registry
 
 
 class AppConfig(PluginConfig):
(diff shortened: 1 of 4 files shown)


netbox_branching/__init__.py
```diff
@@ -18,7 +19,14 @@ class AppConfig(PluginConfig):
 
     def ready(self):
         super().ready()
-        from . import events, search, signal_receivers
+        from . import constants, events, search, signal_receivers
```
arthanson, line 22: I don't think signal_receivers is used

arthanson (approved):

netbox_branching/__init__.py
jeremystretch: This module needs to be imported to register the signal receivers (`record_change_diff()` and `validate_branch_deletion()`).

PR #63, opened 2024-09-02
Closes #52: Introduce the `max_branches` config parameter

### LINKED ISSUE #52: Add a configuration parameter to limit the total number of branches
### Plugin Version

v0.3.0

### Proposed functionality

Introduce a new plugin configuration parameter (e.g. `max_branches`) to set the maximum number of branches that can be created.

### Use case

This will provide a mechanism to guard against the erroneous or malicious creation of an excessive number of branches.

### External dependencies

None

jeremystretch:
### Closes: #52

- Introduce the `max_branches` configuration parameter to limit the total number of branches
- Add a page in the documentation for configuration parameters

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,7 +4,7 @@
 
 Default: None
 
-The maximum number of branches that can be provisioned simultaneously. This includes merged branches that have not been deleted.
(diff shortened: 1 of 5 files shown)


arthanson (approved):
It might be useful in the docs to state why you might want to use this parameter.  As a guess I was thinking there was some performance issue with a lot of branches, but then reading the original issue it was more for security, so it might be worth mentioning that in the docs.

PR #67, opened 2024-09-04
Fixes #66: Capture object representation on ChangeDiff when creating a new object within a branch

### LINKED ISSUE #66: Object representation is not captured when creating a new object within a branch
### Plugin Version

v0.3.1

### NetBox Version

v4.1.0

### Python Version

3.10

### Steps to Reproduce

1. Create and activate a branch
2. Create a new object, such as a site
3. View the branch's diff summary

### Expected Behavior

The creation record for the site should show the associated object's string representation (e.g. the site's name).

### Observed Behavior

The object column shows a string value of "None."

jeremystretch:
### Fixes: #66

Assign the object itself to the ChangeDiff, rather than just its type & ID, to ensure the instance can be cast to a string for the `object_repr` field.

### DIFF
--- a/netbox_branching/signal_receivers.py
+++ b/netbox_branching/signal_receivers.py
@@ -75,8 +75,7 @@ def record_change_diff(instance, **kwargs):
                 current_data = serialize_object(obj, exclude=['created', 'last_updated'])
             diff = ChangeDiff(
                 branch=branch,
-                object_type=instance.changed_object_type,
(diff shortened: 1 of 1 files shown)


arthanson (approved):
Assuming this is the display in question, approved.  Note: You have Ruff check_annotations in other files, looks like a mismatch between the linter and pre-commit?
Change Diffs | NetBox 2024-09-04 07-33-56

Issue #61, opened 2024-08-23
Dry run sync on a branch results in a ValidationError when the branch is later synced with "commit"


mrmrcoleman:
### Plugin Version

0.3.1

### NetBox Version

v4.1-beta1

### Python Version

Python 3.12.3

### Steps to Reproduce

1. Create a branch
2. Create a site in main so that it shows up in the "changes behind" in the branch
3. Sync the branch in dry-run mode
4. Sync the branch with "commit" selected (non dry-run)

### Expected Behavior

Step 4 should work correctly as there are no conflicts between main and the branch

### Observed Behavior

ValidationError({'id': ['Site with this ID already exists.'], 'name': ['Site with this Name already exists.'], 'slug': ['Site with this Slug already exists.']})

PR #70, opened 2024-09-04
#61: Address a regression in the original fix

jeremystretch:
### Fixes: #61

- Revert the change in #65 
- Pass the branch's connection to `transaction.atomic()` to ensure transaction support


arthanson (approved):
It works in the original repro scenario, but it also works in the original code as it has the fix for 61 in it.  So it seems to work, just not sure the fail scenario.

PR #100, opened 2024-09-09
Closes #84: Introduce the `max_active_branches` config parameter

### LINKED ISSUE #84: Add a config parameter to limit the number of active branches
### Plugin Version

v0.4.0

### Proposed functionality

Introduce a new configuration parameter (e.g. `max_active_branches`) to limit the permitted number of active (non-merged) branches. This would complement the `max_branches` parameter introduced in v0.4.0, which limits the _total_ number of branches.

It might also make sense to introduce a per-user variant of this parameter.

### Use case

It may be desirable to limit the number of active branches, either globally or on a per-user basis, to avoid the build-up of abandoned work. This parameter would allow an administrator to enforce a limit on active branches without constraining the total number of branches in the database.

### External dependencies

_No response_

jeremystretch:
### Fixes: #84

- Introduce the `max_active_branches` config parameter
- Enforce it upon the creation of each new Branch

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -144,8 +144,8 @@ class Branch(JobsMixin, PrimaryModel):
             if working_branch_count >= max_working_branches:
                 raise ValidationError(
                     _(
-                        "The configured maximum number of active branches ({max}) cannot be exceeded. One or more "
-                        "active branches must be archived or deleted before a new branch may be created."
+                        "The configured maximum number of working branches ({max}) cannot be exceeded. One or more "
+                        "working branches must be merged or archived before a new branch may be created."
                     ).format(max=max_working_branches)
                 )
 
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -127,15 +127,26 @@ class Branch(JobsMixin, PrimaryModel):
 
     def clean(self):
 
-        # Check whether we're exceeding the maximum number of Branches
+        # Enforce the maximum number of total branches
         if not self.pk and (max_branches := get_plugin_config('netbox_branching', 'max_branches')):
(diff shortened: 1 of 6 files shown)


arthanson (changes requested):
The code works but I think the nomenclature is confusing.  In the branches table there is an "is_active" flag and it is used to denote which branch is currently active, but here we are using active to denote non merged branches. 

If this was set and I got the error the first thing I'd do is go look at my branches table to see which ones are not needed, but I see I don't have any active branches (or at most 1) so I'd be like why did I get this message...  I think the table should easily show which branches are in the state that this flag refers to.

Branches | NetBox 2024-09-10 08-53-25

jeremystretch:
@arthanson thanks that's a really good catch. Maybe we should refer to these branches as "operational" or "provisioned" rather than "active."

jeremystretch:
I settled on `max_working_branches` as the config parameter name.

netbox_branching/models/branches.py
```diff
@@ -127,17 +127,28 @@ def synced_time(self):
...
                     ).format(max=max_branches)
                 )
 
+        # Enforce the maximum number of active branches
+        if not self.pk and (max_working_branches := get_plugin_config('netbox_branching', 'max_working_branches')):
+            working_branch_count = Branch.objects.filter(status__in=BranchStatusChoices.WORKING).count()
+            if working_branch_count >= max_working_branches:
+                raise ValidationError(
+                    _(
+                        "The configured maximum number of active branches ({max}) cannot be exceeded. One or more "
```
arthanson: wording needs to be updated as well

arthanson (changes requested):

arthanson (approved):

PR #109, opened 2024-09-10
Closes #97: Introduce the `exempt_models` config parameter

### LINKED ISSUE #97: Provide a mechanism for other plugins to opt their models out of branching support
### Plugin Version

v0.4.0

### Proposed functionality

Provide a means for a plugin to exclude one or more of its models from branching support, even if they support change logging. This could be one of the following:

- A class from which the model inherits
- An attribute set on the model
- Designation within a `netbox-branching` configuration parameter

Further research is needed to determine which of these is the optimal choice.

### Use case

It may be desirable to exclude certain models within a plugin from branching, even if they support change logging.

### External dependencies

None

jeremystretch:
### Fixes: #97

- Introduce the `exempt_models` config parameter
- Change the `EXCLUDED_APPS` constant to `EXEMPT_MODELS`
- Tweak plugin initialization logic to inspect both static & configured exempt models

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,5 +1,30 @@
 # Configuration Parameters
 
+## `exempt_models`
+
+Default: `[]` (empty list)
+
+A list of models provided by other plugins which should be exempt from branching support. (Only models which support change logging need be listed; all other models are ineligible for branching support.)
+
+Models must be specified by app label and model name, as such:
+
+```python
+exempt_models = (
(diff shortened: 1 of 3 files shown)


arthanson (approved):
Should mention in the docs that the branching plugin must come after any other plugins for the branching support to be enabled for them.

jeremystretch:
@arthanson that's something to be tackled in a separate issue as it's not specific to this FR.

Issue #83, opened 2024-09-06
Add a "share" button under object views when a branch is active


jeremystretch:
### Plugin Version

v0.4.0

### Proposed functionality

Include a button linking to the current object with the current branch active at the top of the object view. This link can be copied and sent to another user to access the same view.

### Use case

As the branch ID does not persist in the URL between page loads, copying the current URL while viewing an object does not capture any information about the current active branch (if any). Providing a branch-aware link to the object provides a convenient mechanism for sharing the current view.

PR #110, opened 2024-09-10
Fixes: #83 - Add Share button for object views when a branch is selected

bctiemann:
### Fixes: #83

Adds a Share button on all object detail pages when a branch other than Main is selected. Button color and placement is TBD (seems like it should be grouped with Bookmark).

Screenshot 2024-09-10 at 6 22 11 PM

### DIFF
--- a/netbox_branching/templates/netbox_branching/inc/share_button.html
+++ b/netbox_branching/templates/netbox_branching/inc/share_button.html
@@ -1,6 +1,6 @@
 {% if active_branch %}
     {% load i18n %}
-    <a href="{{ url }}?_branch={{ active_branch.schema_id }}" type="button" class="btn btn-outline-blue">
+    <a href="{{ object.get_absolute_url }}?_branch={{ active_branch.schema_id }}" type="button" class="btn btn-outline-blue">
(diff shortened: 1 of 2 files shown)


netbox_branching/templates/netbox_branching/inc/share_button.html
```diff
@@ -0,0 +1,6 @@
+{% if active_branch %}
+    {% load i18n %}
+    <a href="{{ url }}?_branch={{ active_branch.schema_id }}" type="button" class="btn btn-outline-blue">
```
jeremystretch: `url` isn't passed as a context variable. The rendered link _does_ work as the browser affixes the query string to the current URL automatically, but in the interest of cleanliness we should either omit `url` or replace it with a call to the object's `get_absolute_url()` method.

jeremystretch: We should add a (translated) title text to the link explaining its purpose (e.g. "Link to this object in the current branch"). This could potentially replace the button's text to save space.

jeremystretch (changes requested):

jeremystretch (approved):

PR #124, opened 2024-09-13
Closes #122: Implement a mechanism to enforce policy before executing branch actions

### LINKED ISSUE #122: Provide a mechanism to configure validators for branch actions (sync, merge, and revert)
### Plugin Version

v0.5.0

### Proposed functionality

Enable the configuration of standalone validators to determine whether a particular branch can have some action performed (i.e. be synced, merged, or reverted). Before any of these operations takes place, the configured validator(s), if any, will be run, with the branch passed as an argument. If any validator returns False, the action will not be permitted.

A validator might look something like this:

```python
def merge_is_allowed(branch):
    # No changes on Friday!
    return datetime.datetime.today().weekday() != 4
```

This would be configured as:

```python
{
    "branch_merge_validators": ["myplugin.utils.merge_is_allowed"]
}
```

### Use case

This feature provides a hook for administrators to enforce custom policies concerning branch actions. For example, an admin might wish to disable merging a branch on certain days (as in the example above), or may require that a branch be logged in some external ticketing system.

### External dependencies

None

jeremystretch:
### Fixes: #122

- Introduce configuration parameters for branch action validators
- Add `can_$action()` methods on the Branch model
- Prevent branch actions from being committed where not permitted by policy
- Dry runs for sync, merge, and revert are still allowed

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -51,3 +51,35 @@ Default: `branch_`
 The string to prefix to the unique branch ID when provisioning the PostgreSQL schema for a branch. Per [the PostgreSQL documentation](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), this string must begin with a letter or underscore.
 
 Note that a valid prefix is required, as the randomly-generated branch ID alone may begin with a digit, which would not qualify as a valid schema name.
+
+---
+
+## `sync_validators`
+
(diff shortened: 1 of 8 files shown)


arthanson (approved):
There should be tests added for these, but as there isn't a mock plugin in the test suite I'm assuming it will be for later.  Maybe add a FR for creating such a test?

Issue #98, opened 2024-09-09
Exception and invalid CablePath state after editing a cable in a branch


Supermathie:
### Plugin Version

0.4.0

### NetBox Version

4.1.0

### Python Version

3.12

### Steps to Reproduce

I'm using the netbox-community docker image.

Create a new admin with `python manage.py createsuperuser`
Provision a new Site: `site`
Create a new Device Role: `test`
Create a new Manufacturer: `Arista`
Load Device Type from devicetype-library: [`AWE-5310.yaml`](https://raw.githubusercontent.com/netbox-community/devicetype-library/refs/heads/master/device-types/Arista/AWE-5310.yaml)
Create 4 Devices: router0[1-4]
Create connection from router01 Ethernet1/1 to router02 Ethernet1/1
Create connection from router01 Ethernet1/2 to router02 Ethernet1/2

![image](https://github.com/user-attachments/assets/a31d4e61-844f-4bd2-9d70-b1158966aed5)

Check Cable #1 trace

![image](https://github.com/user-attachments/assets/857d665d-51cf-42ce-b603-35e077ae8188)

Create branch: `test`
Switch to branch: `test`
Edit Cable #1, switch B side to router03 Ethernet1/1

![image](https://github.com/user-attachments/assets/a8305b5b-c025-4355-8342-35d9a12b50b7)

### Expected Behavior

The edit should complete normally without affecting the Main branch.

### Observed Behavior

Switch to branch: 'Main'

![image](https://github.com/user-attachments/assets/d4bba6a5-dc4a-4ec9-a48b-2b0ae7204162)

Note Cable #1 connection has been affected by the branch edit.

Check Cable #1 trace

![image](https://github.com/user-attachments/assets/85e3c1fe-f3f7-45c5-926c-e0bf0ed38354)

Supermathie:
FYI this is on a test instance (copy of production data) so I am free to try different things.

jeremystretch:
@Supermathie please edit your initial post above to provide exact, step-by-step instructions for reproducing the error. Screenshots are not sufficient. You'll also need to reproduce the issue with a clean v4.1.0 database, as we cannot address potential issues with the original data.

Supermathie:
OK, done.

jeremystretch:
This appears to be a bug in the `save()` method of the Cable model. Calls to `save()` and `delete()` within that method do not respect the active connection and need to be corrected.

VintageCake:
[...]

PR #164, opened 2024-11-01
Fixes #163: Ensure changelog records for non-branching models are created in main schema

### LINKED ISSUE #163: Excluded Models create ObjectChange records in the branch
### Plugin Version

v0.5.1

### NetBox Version

v4.1.5

### Python Version

3.11.9

### Steps to Reproduce

1. Make a new branch
2. Switch to the branch
3. Create a custom field (can do it on circuits, doesn't matter params)
4. merge the branch
5. refresh when merged and see error
6. branch won’t merge as it’s trying to create the custom field in main which already exists.

### Expected Behavior

branch is merged and ObjectChange records for custom fields should be created in main branch as they are in excluded models.

### Observed Behavior

ObjectChange is created in new branch, not main.  When merging it will get an error of duplicate records.

jeremystretch:
### Fixes: #163

- Move model registration to a utility function (housekeeping)
- Revert the removal of database router logic from the fix for #98 (fixes this bug)
- Declare branching support for the CablePath model (alternate fix for #98)

### DIFF
--- a/netbox_branching/database.py
+++ b/netbox_branching/database.py
@@ -21,6 +21,11 @@ class BranchAwareRouter:
             warnings.warn(f"Routing database query for {model} before branching support is initialized.")
             return
 
+        # Bail if the model does not support branching
+        app_label, model_name = model._meta.label.lower().split('.')
+        if model_name not in registry['model_features']['branching'].get(app_label, []):
+            return
+
         # Return the schema for the active branch (if any)
         if branch := active_branch.get():
             return f'schema_{branch.schema_name}'
--- a/netbox_branching/__init__.py
+++ b/netbox_branching/__init__.py
@@ -1,8 +1,8 @@
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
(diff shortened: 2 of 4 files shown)


netbox_branching/database.py
```diff
@@ -21,6 +21,11 @@ def _get_db(self, model, **hints):
             warnings.warn(f"Routing database query for {model} before branching support is initialized.")
             return
 
+        # Bail if the model does not support branching
```
arthanson, line 24: Not for this PR, but might be nice in the future to make this a utility function: model_supports_branching - I could see it might be useful in the future to potentially check elsewhere and it isolates the logic in a nice place / makes it more readable.

arthanson (approved):

netbox_branching/database.py
peteeckel: +1. and it would definitely be useful if a plugin could declare its models branching-aware (or not). That way I could avoid having to tell users it's either branching or NetBox DNS at the moment.

Issue #87, opened 2024-09-06
The current branch (if any) should be deactivated when attempting to create a new branch


jeremystretch:
### Plugin Version

v0.4.0

### NetBox Version

v4.1.0

### Python Version

3.10

### Steps to Reproduce

1. Activate a branch
2. Select "Create a branch" from the branch selector dropdown at top right
3. Enter a branch name and click "Create"

### Expected Behavior

The new branch should be created.

### Observed Behavior

An error is raised:

> Cannot create or modify a branch while a branch is active.

After deactivating the current branch, creating a new branch succeeds.

jeremystretch:
A similar problem arises when attempting to delete a branch while it or another branch is active. I think ideally the controls do so would be disabled while a branch is active, but that may provide infeasible to implement consistently across the UI.

PR #185, opened 2024-12-03
87 deactivate the current branch when creating a new one

arthanson:
### Fixes: #87

I wasn't sure if this was overkill, it looks like we could just always switch the root branch when modifying but wanted to keep this just for creation.


jeremystretch (changes requested):
I worry about silently deactivating the branch, as it may be confusing to the user. IMO we should log a message to the user indicating that the branch has been deactivated.

jeremystretch:
Closing this in favor of #208 (which removes the error messages).

PR #202, opened 2025-01-27
Closes #201: Branch cloning

### LINKED ISSUE #201: Enable branch clonin with point-in-time replay
### Plugin Version

v0.5.2

### Proposed functionality

Implement the ability to clone an unmerged branch by replicating main and replaying the existing  branch's changelog on the new branch.

In the event of an error (e.g. a unique constraint violation) during the replay, it should pause at the last successful change applied. This will leave the new branch in a partially-cloned state, wherein the user can create, modify, and delete objects as with a normal branch. The user should also have the option to resume replaying the changelog from the original branch at any time.

Once the changelog from the original branch has been fully applied to the new branch, the user should have the option to replace the original branch with the cloned branch.

### Use case

Cloning a branch will provide a path for users to resolve issues such as the one captured in bug #32, which cannot be remedied by altering objects within the original branch as the changelog is always replayed in chronological order.

### External dependencies

N/A

jeremystretch:
### Closes: #201

Implements the ability to pull changes from one branch into another by replaying the change records from the source branch onto the destination.

- Add `related_branch` ForeignKey on `BranchEvent` model
- Implements a `pull()` method on the Branch model to effect the replaying of changes
- Introduces a view & REST API endpoint for "pull" operations (consistent with sync & merge operations)
- Introduces the `pre_pull` & `post_pull` signals
- Add `pull_validators` configuration parameter

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -86,6 +86,7 @@ class Branch(JobsMixin, PrimaryModel):
 
     _preaction_validators = {
         'sync': set(),
+        'pull': set(),
         'merge': set(),
         'revert': set(),
         'archive': set(),


arthanson (changes requested):
I think there is a bug here - if you do pull it does not pick up synced changes from the other branch.  After noticing this I use the following test and ran into a error:

1. create branch A
2. create branch B
3. in main create a new site "site-main"
4. activate branch A
5. sync branch A (so it picks up "nsite-main")
6. create a device on branch A using "nsite-main"
7. switch to Branch B
8. Pull from Branch A
9. Error.

I think there are other errors as when I created the device in A it auto-created some power ports from the device type I created, the error seems to be coming from those:

```
{
    "log": [
        "Pulling changes from branch aaa into bbb",
        "Found 10 changes to pull",
        "Applying change 16: tdevaaa created by ahanson using schema_branch_uardeem4",
        "Creating device <DeserializedObject: dcim.Device(pk=107)>",
        "Applying change 17: Input created by ahanson using schema_branch_uardeem4",
        "Creating power port <DeserializedObject: dcim.PowerPort(pk=82)>",
        "{'device': ['device instance with id 107 does not exist.']}"
    ]
}
```

The device I created (using demo data) was:
Name: tdevaaa
Role: Access Switch
Device type: AP7901
Site: site-main

Issue #204, opened 2025-01-29
Support auth_required=True on PluginMenuItem


bctiemann:
### Plugin Version

0.5.2

### Proposed functionality

Once support for it is present in NetBox (v4.2.3), this plugin should add `auth_required=True` to its `PluginMenuItem`s.

### Use case

This will prevent those menu items from showing up in the nav menu if `LOGIN_REQUIRED = False`.

PR #205, opened 2025-01-29
Closes: #204 - Add auth_required to PluginMenuItems

bctiemann:
### Closes: #204

This requires NetBox v4.2.3.

Allows this plugin's menu items to be omitted from the nav menu if `LOGIN_REQUIRED = False`.


jeremystretch (changes requested):
This needs to assert v4.2.3 as the minimum required NetBox version under the PluginConfig.

PR #216, opened 2025-02-11
Replace `pycodestyle` with `ruff`

jeremystretch:
- Swap out `pycodestyle` with `ruff` for linting
- Miscellaneous format cleanup


jeremystretch:
I'd prefer to keep them in scope as we do for NetBox, to ensure that the auto-generated migrations are wrapped to a reasonable line length for legibility. (These can be formatted automatically by running `ruff format migrations/`.)

PR #231, opened 2025-03-11
Fixes #227: Fix scripts trigged via an event rule when the branching plugin is installed

### LINKED ISSUE #227: Script triggered by event rule crashes get_active_branch()
### Plugin Version

0.5.3

### NetBox Version

v4.2.4

### Python Version

3.12.2

### Steps to Reproduce

1. Create an event rule that triggers a Script
2. Trigger the Event Rule

### Expected Behavior

The script should run successfully.

### Observed Behavior

The Script job fails with the error `AttributeError("'NoneType' object has no attribute 'GET'")`.

It seems the line that fails is [here](https://github.com/netboxlabs/netbox-branching/blob/fe76de009d40bf096a280ddb35df81bf943e8527/netbox_branching/utilities.py#L261) and that when the script is called by an Event Rule, the `request` variable is `None`, instead of an instance of `utilities.request.NetBoxFakeRequest` which you see when calling the script normally.

jeremystretch:
### Fixes: #227

ActiveBranchContextManager should attempt to activate a branch only if the request is available.

### DIFF
--- a/netbox_branching/utilities.py
+++ b/netbox_branching/utilities.py
@@ -258,6 +258,6 @@ def ActiveBranchContextManager(request):
     """
     Activate a branch if indicated by the request.
     """
-    if branch := get_active_branch(request):
+    if request and (branch := get_active_branch(request)):
         return activate_branch(branch)
     return nullcontext()


jnovinger (approved):
Sorry it took so long to review such a simple change--it was only my second time working with netbox-branching.

Everything looks good!

PR #236, opened 2025-03-18
Fixes #169: Replicate the global search table for each branch

### LINKED ISSUE #169: Device deleted in branch is not found in main global search
### Plugin Version

0.5.2

### NetBox Version

4.1.6

### Python Version

3.10

### Steps to Reproduce

1. Create a branch
2. Change to created branch
3. delete a device in the test branch
4. Switch back to main
5. Search for the device that was deleted in the test branch
6. The global search will not find the device, but it still shows in the device list.

This issue is also impacting if a device name is changed, as the value is changed across all branches. Example:

1. Go to test branch
2. Change name of a device
3. Switch back to main
4. Search with original name
5. Device is not found with original name as the 'value' updates to the new name across all branches.
6. The device with the original name can still be found in the devices page without using the global search.

### Expected Behavior

Device changes is done only on the test branch, with the main branch being left untouched until merge

### Observed Behavior

Changes were done across all branches

jeremystretch:
### Fixes: #169

- Replicate the global search table for each branch
- Modifications to objects performed within a branch are reflected in the branch's local search table
- Searching with a branch active queries its local copy of the search table

### DIFF
--- a/netbox_branching/constants.py
+++ b/netbox_branching/constants.py
@@ -14,6 +14,7 @@ QUERY_PARAM = '_branch'
 # must be replicated for each branch to ensure proper functionality
 INCLUDE_MODELS = (
     'dcim.cablepath',
+    'extras.cachedvalue',
 )
 
 # Models for which branching support is explicitly disabled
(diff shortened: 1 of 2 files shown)


jnovinger (approved):
Checked all the paths mentioned in the original and any others I could think of. This looks and feels solid!

PR #263, opened 2025-05-20
Closes #262: Copy migrations table to branch

### LINKED ISSUE #262: Copy current migrations table when provisioning a new branch
### Plugin Version

v0.5.5

### Proposed functionality

When a new branch is provisioned, copy the `django_migrations` database table to maintain a record of migrations that have been applied.

### Use case

This is necessary to implement FR #1, which is slated for v0.6.0. It needs to be implemented _prior_ to the v0.6.0 release to ensure that migration tables can be replicated to existing branches before upgrading to NetBox v4.3.

### External dependencies

N/A

jeremystretch:
### Closes: #262

- Extends the `provision()` method on Branch to replicate the current state of the `django_migrations` table
- Includes a migration to copy the `django_migrations` tables to all existing branches


netbox_branching/models/branches.py
```diff
@@ -540,6 +540,29 @@ def provision(self, user):
                     f"ALTER TABLE {schema_table} ALTER COLUMN id SET DEFAULT nextval(%s)", [sequence_name]
                 )
 
+                # Copy the migrations table
+                main_table = f'{MAIN_SCHEMA}.django_migrations'
```
jnovinger, line 544: Might be worth noting this needs to change in v0.6.0, since the other references to `MAIN_SCHEMA` are already changed in #254.

jnovinger (commented):
Sorry it took a bit, got myself confused between the new `main_schema` plugin parameter coming in v0.6.0 with the replaced `MAIN_SCHEMA` constant and where this was landing relative to that. 🤦 

I was verifying that all migrations were copied as part of `0004_copy_migrations` and realized that, pretty much by definition, it can't copy itself--so all existing branches that this migration updates will be behind by this single migration. I can't decide if I think that's a problem or not.

Just had one other note to leave yourself a reminder, but looks good otherwise.

PR #264, opened 2025-05-21
Closes #1: Enable branch migrations

### LINKED ISSUE #1: Enable database migrations with open branches
### Plugin Version

v0.2.0

### Proposed functionality

Implement the ability to apply database migrations (e.g. from a NetBox upgrade) with one or more open branches. Currently, this is not supported as migration operations are applied only to the main schema.

### Use case

This would enable users to reliably upgrade to new NetBox versions without needing to first merge any open branches.

### External dependencies

_No response_

jeremystretch:
### Closes: #1

- Introduce the "migrating" state & associated signals for branches
- Add a BranchEvent type for migration
- Add the `applied_migrations` ArrayField on Branch to record schema migrations which have been applied to the branch
- Override the `allow_migrate()` method on our custom database router to suppress migrations of models which do not support branching
- Introduce the `migrate()` method on Branch & add supporting UI view
- The `apply()` and `undo()` methods on our proxy model for ObjectChange now take a Branch instance as the first argument
- Add a `migrate()` method on ObjectChange to automatically adapt pre- and post-change data according to the applicable migrators


jeremystretch:
> 1. Any branch created before having branching 0.5.6 cannot be migrated?

Any branch can be migrated, but v0.5.6 of the plugin must be installed _before_ upgrading NetBox.

> 2. If you are on an older version of NetBox 4.1.9 and you upgrade branching to 0.5.6 then you upgrade NetBox to 4.2 and make changes on the branch, then update to NetBox 4.3 and more changes to the branch then upgrade to this PR and do the branch migration - should that still be valid?

Yes, that should be supported. However, NetBox v4.3.2 or later is necessary to enable migrating branches.

> 3. The branch in the list shows as Ready before you run the branch migration, it also shows in the dropdown as "{name} (Ready)" but grayed - should we make that show a different status (Pending Migration) or something as it isn't really ready?

Its status is technically "ready" however the branch is not available for use due pending migrations. We could introduce an explicit status for this, however I'm not sure how we'd reliably update the branch's status to it.

jeremystretch:
Still working on this. Want to see if it's feasible to add a discrete "needs migration" branch status to improve UX.

docs/models/branch.md
```diff
@@ -20,17 +20,22 @@ The unique, randomly-generated identifier of the PostgreSQL schema which houses
...
+| New          | Not yet provisioned in the database                                |
+| Provisioning | A job is running to provision the branch's PostgreSQL schema       |
+| Ready        | The branch is healthy and ready to be synchronized or merged       |
+| Syncing      | A job is running to synchronize changes from main into the branch  |
+| Migrating    | A job is running to apply database migrations to the branch schema |
+| Merging      | A job is running to merge changes from the branch into main        |
+| Reverting    | A job is running to revert previously merged changes in main       |
+| Merged       | Changes from this branch have been successfully merged into main   |
+| Archived     | A merged branch which has been deprovisioned in the database       |
+| Failed       | Provisioning the schema for this branch has failed                 |
```
arthanson: Needs Pending Migrations added here

PR #274, opened 2025-06-04
Fixes #256: Show changes ahead/behind only for "ready" branches

### LINKED ISSUE #256: New branches show changes ahead until Ready
### Plugin Version

0.5.4

### NetBox Version

4.2.3

### Python Version

Python 3.12.3

### Steps to Reproduce

1. Install both assurance and branching
2. Send X entities to assurance and either apply them or ignore them
3. Create a new branch

### Expected Behavior

The branch shows 0 `Changes Ahead`, until some changes are added to the branch

### Observed Behavior

* When in the `Provisioning` state, the branch shows X-1 `Changes Ahead`
* When the branch enters the `Ready` state, the branch returns to correctly showing 0 `Changes Ahead`

jeremystretch:
### Fixes: #256

Compute changes ahead & behind only for branches in the "ready" state

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -204,32 +204,32 @@ class Branch(JobsMixin, PrimaryModel):
         """
         Return a queryset of all ObjectChange records created in main since the Branch was last synced or created.
         """
-        if self.status not in BranchStatusChoices.WORKING:
-            return ObjectChange.objects.none()
-        return ObjectChange.objects.using(DEFAULT_DB_ALIAS).exclude(
-            application__branch=self
-        ).filter(
-            changed_object_type__in=get_branchable_object_types(),
-            time__gt=self.synced_time
-        )
+        if self.status == BranchStatusChoices.READY:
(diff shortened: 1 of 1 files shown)


netbox_branching/models/branches.py
```diff
@@ -204,32 +204,32 @@ def get_unsynced_changes(self):
...
     def get_merged_changes(self):
         """
         Return a queryset of all merged ObjectChange records for the Branch.
         """
-        if self.status not in (BranchStatusChoices.MERGED, BranchStatusChoices.ARCHIVED):
-            return ObjectChange.objects.none()
-        return ObjectChange.objects.using(DEFAULT_DB_ALIAS).filter(
-            application__branch=self
-        )
+        if self.status in (BranchStatusChoices.MERGED, BranchStatusChoices.ARCHIVED):
```
jeremystretch, line 228: This change wasn't strictly necessary but I flipped the logic to be consistent with the other two methods.

jeremystretch:
I didn't include the failed state because it only occurs during initial provisioning. That may change in the future but for now I think it's correct.

PR #276, opened 2025-06-04
Fixes #275: Record last_sync time when provisionin a branch

### LINKED ISSUE #275: Branch sync time is not set during initial provisioning
### Plugin Version

v0.5.5

### NetBox Version

v4.3.1

### Python Version

3.10

### Steps to Reproduce

1. Disable all background worker processes (`manage.py rqworker`) to simulate a delay in branch provisioning.
2. Create a new branch. Its status should read "new" as it has not yet been provisioned.
3. Create a site (or any object) in main.
4. View the branch.

### Expected Behavior

The "last synced" time for the branch should be null, as the branch has not yet been provisioned.

### Observed Behavior

The "last synced" time indicated the time at which the branch object was created in the database. This is misleading, because any changes made to main between the time when the branch is created to when it is actually provisioned will be captured in the branch.

jeremystretch:
### Fixes: #275

Note: This should be rebased once PR #274 is merged into main, as it will introduce merge conflicts.

- Set the `last_sync` time when provisioning a branch
- Remove the `synced_time` property from Branch

### DIFF
--- a/netbox_branching/templates/netbox_branching/branch.html
+++ b/netbox_branching/templates/netbox_branching/branch.html
@@ -83,13 +83,17 @@
           <tr>
             <th scope="row">{% trans "Last synced" %}</th>
             <td>
-              {{ object.synced_time|isodatetime }}
-              {% if object.is_stale %}
-                <span class="text-danger" title="{% trans "Branch is stale and can no longer be synced" %}">
-                  <i class="mdi mdi-alert-circle"></i>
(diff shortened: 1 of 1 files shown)


arthanson (changes requested):
It's now null but now it showing just "ago" which is confusing
Monosnap nbx | NetBox 2025-06-06 08-26-20

PR #329, opened 2025-09-09
Fixes #150: Fix recalculation of cable paths when merging a branch with cable changes

### LINKED ISSUE #150: Cable paths are not updated when merging a branch with cable changes
### Plugin Version

v0.5.0

### NetBox Version

v4.1.1

### Python Version

3.10

### Steps to Reproduce

This bug addresses a secondary problem raised in #98.

1. Create two devices, A and B.
2. Create interface 1 on device A and interfaces 1 & 2 on device B.
3. Create a cable between device A interface 1 and device B interface 1.
4. Create and activate a branch.
5. Edit the cable, and change the device B end from interface 1 to interface 2.
6. Merge the branch, committing changes.
7. Inspect the connection status of device A interface 1 and device B interface 2.

### Expected Behavior

Either interface should show as connected to the other.

### Observed Behavior

Although both interfaces show as connected to the same cable, no path is populated to show them as connected to each other.

As a workaround, manually retracing the cables path (`manage.py trace_paths --force`) corrects this condition.

jeremystretch:
### Fixes: #150

**NOTE:** This must be tested against the [`20241-cable-change-records`](https://github.com/netbox-community/netbox/tree/20241-cable-change-records) branch of NetBox, as it depends on changes made therein.

- Call a model's custom `serialize_object()` and `deserialize_object()` methods if defined
- Extend `update_object()` to support assigning non-field attributes

### DIFF
--- a/netbox_branching/__init__.py
+++ b/netbox_branching/__init__.py
@@ -14,7 +14,7 @@ class AppConfig(PluginConfig):
     description = 'A git-like branching implementation for NetBox'
     version = '0.7.0'
     base_url = 'branching'
-    min_version = '4.4.0'
+    min_version = '4.4.1'
     max_version = '4.4.99'
     middleware = [
         'netbox_branching.middleware.BranchMiddleware'


netbox_branching/models/changes.py
```diff
@@ -48,14 +48,18 @@ def apply(self, branch, using=DEFAULT_DB_ALIAS, logger=None):
 
         # Creating a new object
         if self.action == ObjectChangeActionChoices.ACTION_CREATE:
-            instance = deserialize_object(model, self.postchange_data, pk=self.changed_object_id)
+            if hasattr(model, 'deserialize_object'):
```
jnovinger, line 51: Tiny edge-case here, it's possible that some model, somewhere (in core in the future or a plugin model that nominally supports branching) has a `deserialize_object` attr that is not callable.

I'm only calling this out because I could see a random plugin doing something like this and it costing us a fair amount of time to debug and respond.

jnovinger (approved):
Approving this as is, as it fixes the issue. However, I did point out what I thought was a potential edge-case worth handling.

netbox_branching/models/changes.py
jeremystretch: It's a valid concern, but seems very unlikely, especially since this change is limited to the branching plugin. I would like to establish `deserialize_object()` as a documented method on ChangeLoggingMixin alongside `serialize_object()` in NetBox v4.5. That would further reduce the surface area for a potential naming collision.

jeremystretch:
The reason we're only seeing five changes after merging the branch is that the ObjectChange record resulting from the sixth and final change application shows the same pre- and post-change values for the A & B terminations; no ObjectChange record is recorded in main (unless `CHANGELOG_SKIP_EMPTY_CHANGES` is set to True).

I'm not sure there's a reasonable way to work around this, because in this case we can't compare to the instance's _original_ state as we normally would under Cable's `save()` method.

PR #348, opened 2025-10-28
Fixes #342: Error on update form if object has been deleted in main

### LINKED ISSUE #342: Display warning if object has been deleted in main branch
### Plugin Version

v0.7.1

### Proposed functionality

Handle ____DoesNotExist exceptions when updating objects that have been deleted in the base branch. I think we could handle this flow better, my suggestion is a modal that redirects users to sync their branch with the base branch. 

* Create an object in main (like a site)
* Create a branch
* In main delete the site you created
* Go to the branch and select the site
* edit the site (change the description) and save
* Get a server error

### Use case

When making changes in branches to objects that have been deleted in the base branch, we throw this exception. For users that aren't familiar with Django, the server error is intimidating.

### External dependencies

N/A

arthanson:
### Fixes: #342 

Adds a warning if you try to modify an object that has been deleted in main.  It is possible to make a fix to create the changelog without referring to main, however this will still result in an un-mergeable branch because of the way the merge plays back the changelogs in order, so a user message and preventing the save was decided as the better course of operation.

For delete - the post_save signal can come in before the pre_delete so put the check in both places. 

This covers the update/delete of items that originally existed in main and were deleted in main but still exist in the branch. Also checks are made if the object was only created in the branch (so it doesn't exist in main as it never existed in main).

Monosnap Image 2025-10-28 15-07-32

### DIFF
--- a/netbox_branching/signal_receivers.py
+++ b/netbox_branching/signal_receivers.py
@@ -1,6 +1,7 @@
 import logging
 from functools import partial
 
+from django.contrib.contenttypes.models import ContentType
 from django.db import DEFAULT_DB_ALIAS
 from django.db.models.signals import post_migrate, post_save, pre_delete
 from django.dispatch import receiver
@@ -32,6 +33,38 @@ __all__ = (
 )
 
 
+def check_object_accessible_in_branch(branch, model, object_id):
+    """
+    Check if an object is accessible for operations in a branch.
+
+    An object is accessible if it either exists in main or was created in the branch.
+    This prevents operations on objects that were deleted in main.
+
+    Args:
+        branch: The Branch instance
+        model: The model class
+        object_id: The primary key of the object
+
+    Returns:
+        True if the object is accessible (exists in main or was created in branch), False otherwise
+    """
+    with deactivate_branch():
+        try:
+            model.objects.get(pk=object_id)
+            return True
+        except model.DoesNotExist:
+            pass
+
+    # Object doesn't exist in main - check if it was created in the branch
+    content_type = ContentType.objects.get_for_model(model)
+    return ChangeDiff.objects.filter(
+        branch=branch,
+        object_type=content_type,
+        object_id=object_id,
+        action=ObjectChangeActionChoices.ACTION_CREATE
+    ).exists()
+
+
 @receiver(post_clean)
 def validate_branching_operations(sender, instance, **kwargs):
     """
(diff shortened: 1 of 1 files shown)


netbox_branching/signal_receivers.py
```diff
@@ -25,9 +28,43 @@
...
+    Validate that branching operations are valid (e.g., not modifying deleted objects).
+    """
+    branch = active_branch.get()
+
+    # Only validate if we're in a branch and this model supports branching
+    if branch is None:
+        return
+
+    # Check if this model supports branching
+    content_type = ContentType.objects.get_for_model(instance)
```
jnovinger: Any reason to not query on `ObjectType` directly here?

netbox_branching/signal_receivers.py
```diff
@@ -25,9 +28,43 @@
...
+    if 'branching' not in content_type.object_type.features:
+        return
+
+    # For updates and deletes, check if the object exists in main
```
jnovinger: I'm probably missing it, but how are we accounting for deletes?

netbox_branching/signal_receivers.py
```diff
@@ -25,9 +27,46 @@
...
+    branch = active_branch.get()
+
+    # Only validate if we're in a branch and this model supports branching
+    if branch is None:
+        return
+
+    # Check if this model supports branching
+    object_type = ObjectType.objects.get(app_label=instance._meta.app_label, model=instance._meta.model_name)
+    if 'branching' not in object_type.features:
+        return
```
jnovinger, line 85: It might be safer to use `ObjectType.objects.get_for_model()` here, since it wraps `.get_or_create()` and also contains the workaround for that signals/migration bug we saw in v4.4.0.

netbox_branching/signal_receivers.py
```diff
@@ -136,6 +189,46 @@ def handle_branch_event(event_type, branch, user=None, **kwargs):
...
+
+    # Only validate if we're in a branch
+    branch = active_branch.get()
+    if branch is None:
+        return
+
+    # Check if this model supports branching
+    try:
+        object_type = ObjectType.objects.get(app_label=instance._meta.app_label, model=instance._meta.model_name)
```
jnovinger: Same as above

netbox_branching/signal_receivers.py
```diff
@@ -136,6 +189,46 @@ def handle_branch_event(event_type, branch, user=None, **kwargs):
...
+            return
+    except ObjectType.DoesNotExist:
+        return
+
+    # For deletions, check if the object exists in main
+    if hasattr(instance, 'pk') and instance.pk is not None:
+        model = instance.__class__
+        try:
+            # Try to get the object from main database (explicitly use default connection)
+            model.objects.using(DEFAULT_DB_ALIAS).get(pk=instance.pk)
```
jnovinger: It seems like we're doing this checking in `main` differently in this signal receiver than in `validate_branching_operations()`. Is that intentional? If not, then I think we should stick with way or the other to be consistent.

jnovinger (changes requested):

arthanson:
@jnovinger There was a bug case where if you created the object in the branch it was still complaining that it wasn't existing in main.  I refactored the code to a common function as the testing was more complicated.

jnovinger (approved):

PR #364, opened 2025-11-11
Fixes #358: Close branch connections after CONN_MAX_AGE expires

### LINKED ISSUE #358: Database connections increase over time, leading to errors
### Plugin Version

0.7.1

### NetBox Version

4.4.5

### Python Version

3.11

### Steps to Reproduce

When branching is used extensively (many branches), we observe the connection count to the PSQL database increasing with each branch that is provisioned, and every time you click into a branch. Over time, this causes the connection count to max out,  no new connections can be opened, and NetBox throws errors.

This was first seen from a customer in NBE, however I've reproduced on the latest community and branching versions (4.4.5 and 0.7.1).

To reproduce:

open a bunch of branches (took me about 40)

open each branch in the GUI

observe the connection count increase to postgres (select count(*) from pg_stat_activity;)

around 105 connections with default settings, it will prevent new connections

Image
Image

### Expected Behavior

DB connections should close automatically when they are no longer in use.

### Observed Behavior

DB connections remain open, increasing linearly until the max connection limit is reached and errors occur.

jnovinger:
### Fixes: #358

Branch-specific database connections never close after `CONN_MAX_AGE` expires, causing connection exhaustion when users visit multiple branches. Django's `close_old_connections()` iterates `DATABASES.keys()` to discover aliases/names for connections to cleanup, but branch aliases are served dynamically via `DynamicSchemaDict.__getitem__()` without being yielded by `__iter__()`, so they're never checked for expiry. They're essentially invisible to Django's `CONN_MAX_AGE` machinery.                                                                                                                                                                                     
                                                                                                                                                                                                
Solution Approaches                                                                                                                                                                             
                                                                                                                                                                                                
1. Caching aliases in `DATABASES` dict: Fixed cleanup but broke all 43 existing tests with "Database connections not allowed in this test." Django's test framework checks `DATABASES.keys()` during setup and blocks non-default connections. The issue wasn't just about adding `databases = '__all__'` to existing tests (or to some base test class), which this would have required. The deeper concern was that by making branch aliases visible in `DATABASES.keys()` during test setup, we'd be changing the fundamental isolation guarantees. Tests that never needed to know about branch databases would suddenly have them in scope, potentially causing cross-contamination between tests or allowing tests to accidentally depend on branch state.

2. Custom `__iter__` using private API: Used `connections._connections._storage.__dict__` to discover cached aliases. Fixed cleanup and all tests passed, but relied on undocumented Django internals. Rejected for maintenance concerns.
                                                                                                  
3. (what I ended up with) Track branch aliases in our own thread-local storage when first accessed, register cleanup handlers on `request_started` and `request_finished` signals. Uses only public APIs (`asgiref.Local`, Django signals, `connections.close_if_unusable_or_obsolete`). Matches Django's pattern where `DATABASES.keys()` is static and implements thread-specific tracking in the same manner as Django.

### DIFF
--- a/netbox_branching/tests/test_connection_lifecycle.py
+++ b/netbox_branching/tests/test_connection_lifecycle.py
@@ -11,6 +11,8 @@ from netbox_branching.utilities import activate_branch, close_old_branch_connect
 
 @tag('regression')  # netbox-branching #358
 class BranchConnectionLifecycleTestCase(TransactionTestCase):
+    serialized_rollback = True
+
     def setUp(self):
         """Set up test environment with CONN_MAX_AGE=1."""
         self.original_max_age = settings.DATABASES['default'].get('CONN_MAX_AGE', 0)


mgaruccio:
this may not be worth fixing, but FYI that this seems to only close connections properly if they get initiated via api or UI, if you create a branch with mgmt commands it will leave the connection idle indefinitely.

probably not a problem outside of my testing scenario but could be if there's ever another path to create connections that bypasses the http api?

arthanson (approved):
really nice work @jnovinger

Issue #198, opened 2025-01-22
Implement the ability to squash individual changes when merging a branch


jeremystretch:
### Plugin Version

v0.5.2

### Proposed functionality

When a branch is being merged, enable the user to select specific changes to "squash." Squashing collapses one or more successive operations within a branch into the first operation for the affected object.

For example, consider the follow set of actions performed within a branch:

1. Create a site named "Site 1" with status "Active"
2. Create a device within the site
3. Change the site's status to "Planned"
4. Assign the site to a region
5. Add another device to the site

In this example, changes 1, 3, and 4, can be safely collapsed into a single create operation.

### Use case

The primary driver for this functionality is to enable users to rectify conflicts preventing a branch from being merged. (This issues is largely captured under #32.) For example, consider the following series of events:

1. Branch A is created.
2. A new site named Site 1 is created in Branch A.
3. A new site with the same name is created in the main database.
4. A user attempts to merge Branch A. This fails due to the duplicate site name.
5. The user renames Site 1 to Site 2 within the branch.
6. The user attempts again to merge Branch A. This _still_ fails because the change set is replayed in chronological order: The initial creation of the site will always fail before the following update is processed.

With the ability to squash changes, the update to the site object can be squashed into the original create action, rectifying the conflict and allowing the branch to be merged successfully.

### External dependencies

N/A

PR #370, opened 2025-11-20
Add ability to Squash Changes when merging

arthanson:
### Fixes: #198

This squashes (collapses) all Changelogs (**for a given object**) in a branch to a single Change:

- CREATE + (any updates) + DELETE = skip entirely
- (anything other than CREATE) + DELETE = DELETE
- CREATE + UPDATEs = CREATE
- multiple UPDATEs = UPDATE

The old (iterative) and new (collapse) merge strategy are both supported (the default is the old iterative approach). There is a select-box on merge to enable the squash merge:

Monosnap Merge Branch b1 | NetBox 2025-11-25 17-38-43

Because ObjectChanges can reference other objects (FK) it has to do dependency ordering: If an object references something that is created, the create has to come first, etc..

This will not fix 100% of the issues, but the goal is that the end user can fairly easily fix the branch (by either deleting a problem Object or modifying it's data) and be able to get the branch to merge.  

The biggest thing this fixes is Where you have an Object that causes a Constraint error. (Site with a slug that conflicts with one in main). Currently using the replay of changelogs you cannot fix this as you will always try to replay the ObjectcChange with the conflicting state.  With collapse migrations, the intermediate state of that caused the conflict is skipped.

The major thing this can't fix is the swap case: You have a Module Bay with two modules at position 1 and 2 and you swap these (move 1 to 3, 2 to 1, 3 to 2) as with collapse you miss the move to position 3 and no matter which one you try to merge you will run into a constraint error.  But, you can easily fix this by temporarily modifying the data before merge then merging.

**Note:** I looked at doing partial collapse (i.e. only collapsing some of the changes) but quickly ran into issues because of the constraint issues mentioned above and that Objects can reference other objects which means you have to do a dependency change and at that point it is just cleaner to collapse all the migrations as you have to do all the work anyways and it gets messy figuring out what you can partially collapse.

**Step1:** For collapse you want to take the initial objects prechange_data, and then you want to collapse all the objects postchange_data down into one so later field modifications overwrite older ones.

**Note:** because you are collapsing you sort of loose the time sequence (do you take it from the first change or the last - which actually depends). So barring dependencies we generally do DELETEs, UPDATEs then CREATEs (within those in time order)

**Step 2:** For dependency ordering it matters what type of operation you are doing.  For a create you care about if postchange_data from another object refers to it, but for a delete you care if the prechange_data for another object refers to it. **Note:** we only care if these references are to other objects being modified in this branch.  If it is a reference to something in main we don't need to track the dependency.

**Step 3:** For actually doing the dependency ordering a given Change can reference more then one FK so we use an algorithm that handles this and have to have cycle detection, which hopefully shouldn't happen (A references B, references C, references A) but need to guard against it.

**Note:** There is an edge case with nullable FK fields and having a created item refer to each other (Circuit / CircuitTermiantions apparently do this) If we detect this we split the create back into a create (with the nulled fields) and an update setting those fields.

**Note:** All of these new ObjectChange are in-memory, we don't actually modify the existing ObjectChange in the database. Because of this we can't use the existing apply / revert code as those work on actual ObjectChange records, so we have new functions similar to those.

`_build_fk_dependency_graph` has very similar code for each type of Change, but it is just dissimilar enough I didn't see a good way of refactoring it to be more DRY without increasing the complexity.  I also created some sub-functions which are only called once (like `_build_fk_dependency_graph`) but it encapsulates the given functionality and makes it far easier to understand IMHO.

Right now there is a lot of logging that will appear in the Merge job log making it fairly big. This info is useful for debugging and it was decided to error on the side of too-much info rather than paring it down. We can do that later after it has been out and proven itself.

**Note:** From review changes - updated so squash revert / merge use the requesting user as the user for the ChangeLogs.

### DIFF
--- a/netbox_branching/merge_strategies/strategy.py
+++ b/netbox_branching/merge_strategies/strategy.py
@@ -14,8 +14,8 @@ class MergeStrategy(ABC):
     """
     Abstract base class for merge strategies.
     """
-    # Ordering for changes queryset in merge() and revert() operations
-    merge_changes_ordering = 'time'  # Chronological order (oldest first)
+    # Ordering for changes queryset in revert() operation
+    # Merge always uses chronological order ('time')
     revert_changes_ordering = '-time'  # Reverse chronological order (newest first)
 
     @abstractmethod
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -533,12 +533,8 @@ class Branch(JobsMixin, PrimaryModel):
         # Emit pre-merge signal
         pre_merge.send(sender=self.__class__, branch=self, user=user)
 
-        # Get the merge strategy to determine the correct ordering for changes
-        strategy_class = get_merge_strategy(self.merge_strategy)
-        changes_ordering = strategy_class.merge_changes_ordering
-
         # Retrieve staged changes before we update the Branch's status
-        if changes := self.get_unmerged_changes().order_by(changes_ordering):
+        if changes := self.get_unmerged_changes().order_by('time'):
             logger.info(f"Found {len(changes)} changes to merge")
         else:
             logger.info("No changes found; aborting.")
@@ -557,7 +553,8 @@ class Branch(JobsMixin, PrimaryModel):
 
         try:
             with transaction.atomic():
-                # Execute the merge strategy
+                # Get and execute the appropriate merge strategy
+                strategy_class = get_merge_strategy(self.merge_strategy)
                 logger.debug(f"Merging using {self.merge_strategy} strategy")
                 strategy_class().merge(self, changes, request, logger, user)
 


netbox_branching/models/branches.py
```diff
@@ -96,6 +101,11 @@ class Branch(JobsMixin, PrimaryModel):
         null=True,
         related_name='+'
     )
+    merged_using_collapsed = models.BooleanField(
```
jeremystretch: Propose changing this to a CharField named `merge_strategy` with choices "Iterative" and "Squash." This will enable us to easily add new strategies in the future should the need arise.

netbox_branching/models/branches.py
```diff
@@ -501,6 +512,730 @@ def migration_progress_callback(action, migration=None, fake=False):
 
     migrate.alters_data = True
 
+    # Helper class and functions for collapsing ObjectChanges during merge
+    class CollapsedChange:
```
jeremystretch: Why does this live inside Branch? Seems like it should be a standalone class.

netbox_branching/models/branches.py
```diff
@@ -501,6 +512,730 @@ def migration_progress_callback(action, migration=None, fake=False):
...
+        def __repr__(self):
+            obj_id = self.key[1]
+            suffix = f" ({self.key[2]})" if len(self.key) > 2 else ""
+            return (
+                f"<CollapsedChange {self.model_class.__name__}:{obj_id}{suffix} "
+                f"action={self.final_action} changes={len(self.changes)}>"
+            )
+
+    @staticmethod
+    def _diff_object_change_data(action, prechange_data, postchange_data):
```
jeremystretch: Can we convert all these static methods to standalone utility functions? They don't seem particularly relevant to the branch itself.

netbox_branching/models/branches.py
```diff
@@ -501,6 +512,730 @@ def migration_progress_callback(action, migration=None, fake=False):
...
+            'pre': {
+                k: prechange_data.get(k) for k in changed_attrs
+            },
+            'post': {
+                k: postchange_data.get(k) for k in changed_attrs
+            },
+        }
+
+    @staticmethod
+    def _collapse_changes_for_object(changes, logger):
```
jeremystretch: This seems like it should be a method on CollapsedChange, e.g. `collapse()`.

netbox_branching/migrations/0007_branch_merged_using_collapsed.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 5.2.8 on 2025-11-20 21:30
+
+from django.db import migrations, models
```
jnovinger:
```suggestion
from django.db import migrations, models
```

netbox_branching/tests/test_merge.py
```diff
@@ -0,0 +1,820 @@
...
+        self.user = User.objects.create_user(username='testuser')
+
+        # Create some base objects in main
+        self.manufacturer = Manufacturer.objects.create(name='Manufacturer 1', slug='manufacturer-1')
+        self.device_type = DeviceType.objects.create(
+            manufacturer=self.manufacturer,
+            model='Device Type 1',
+            slug='device-type-1'
+        )
+        self.device_role = DeviceRole.objects.create(name='Device Role 1', slug='device-role-1')
```
jnovinger, line 38: Any chance we can move any of this setup to the `setUpTestData` class-method so that it only runs once? Or is that not possible because of `serialized_rollback`?

jnovinger (commented):
Building on Jeremy's suggestion to make this a CharField with strategy choices: we should take this further and refactor the merge/revert logic to use the actual Strategy pattern.

Right now we've got the boolean check scattered through `merge()` and `revert()` with `_merge_collapsed()` vs `_merge_iterative()` methods. If we're going to support multiple strategies (and "Squash" strongly implies there could be others), we should pull this into proper strategy classes.

Something like:

```python
# strategies.py
class MergeStrategy(ABC):
    @abstractmethod
    def merge(self, branch, changes, request, commit, logger): ...

    @abstractmethod
    def revert(self, branch, changes, request, commit, logger): ...

class IterativeMergeStrategy(MergeStrategy):
    # current _merge_iterative / _revert_iterative logic

class SquashMergeStrategy(MergeStrategy):
    # current _merge_collapsed / _revert_collapsed logic
```

Then `Branch.merge()` becomes:

```python
def merge(self, user, commit=True):
    strategy = get_strategy(self.merge_strategy)  # looks up from field/registry
    # ... setup ...
    strategy.merge(self, changes, request, commit, logger)
    # ... cleanup ...
```

This gets the 800+ lines of merge logic out of the Branch model, makes it trivial to add new strategies, and each strategy is independently testable. Also means `CollapsedChange` and all the dependency ordering helpers move into `SquashMergeStrategy` where they belong. They're implementation details of that specific strategy, not general Branch concerns.

jnovinger:
> Building on Jeremy's suggestion to make this a CharField with strategy choices: we should take this further and refactor the merge/revert logic to use the actual Strategy pattern.
> ...
> This gets the 800+ lines of merge logic out of the Branch model, makes it trivial to add new strategies, and each strategy is independently testable. Also means `CollapsedChange` and all the dependency ordering helpers move into `SquashMergeStrategy` where they belong. They're implementation details of that specific strategy, not general Branch concerns.

One big benefit on this approach is that it makes each of the strategies much more testable at a lower level (e.g. not black-box testing `Branch.merge()`).

Test coverage is great for end-to-end scenarios, but I think we might want to expand verification of the internal dependency ordering logic. All the existing tests verify that merges succeed/fail correctly, which is good, but they don't inspect whether the dependency graph is actually being built correctly.

netbox_branching/tests/test_merge.py
arthanson: I don't think that is possible with TransactionTestCase used because it's creating branches.

netbox_branching/jobs.py
```diff
@@ -122,6 +134,12 @@ class RevertBranchJob(JobRunner):
     """
     class Meta:
         name = 'Revert branch'
+        job_timeout = 3600  # 1 hour - increased for large reverts
+
+    @property
+    def job_timeout(self):
+        """Return the job timeout from Meta."""
+        return getattr(self.Meta, 'job_timeout', None)
```
jnovinger: This seems a bit redundant. Any way to factor it out to a single implementation?

netbox_branching/models/branches.py
```diff
@@ -645,6 +633,7 @@ def revert(self, user, commit=True):
         self.status = BranchStatusChoices.READY
         self.merged_time = None
         self.merged_by = None
+        self.merge_strategy = BranchMergeStrategyChoices.ITERATIVE
```
jnovinger: I'm not sure I understand why we're setting this to `ITERATIVE` here. Can we be sure we always want the iterative strategy?

netbox_branching/collapse_merge.py
```diff
@@ -0,0 +1,675 @@
+"""
```
jnovinger: It seems like this file is left-over from the refactor, yeah?

jnovinger (changes requested):

netbox_branching/models/branches.py
```diff
@@ -96,6 +96,13 @@ class Branch(JobsMixin, PrimaryModel):
         null=True,
         related_name='+'
     )
+    merge_strategy = models.CharField(
```
jeremystretch, line 98: Should this be null for branches that have not yet been merged? `BranchActionForm` appears to default to the iterative strategy regardless of what is set on the branch.

netbox_branching/choices.py
```diff
@@ -46,6 +46,16 @@ class BranchStatusChoices(ChoiceSet):
     )
 
 
+class BranchMergeStrategyChoices(ChoiceSet):
+    ITERATIVE = 'iterative'
+    SQUASH = 'squash'
+
+    CHOICES = (
+        (ITERATIVE, _('Iterative')),
+        (SQUASH, _('Squash')),
```
jeremystretch: Providing context for translators so we don't end up with vegetable names.
```suggestion
        (SQUASH, _('Squash', 'The act of compressing multiple records into one')),
```

netbox_branching/forms/misc.py
```diff
@@ -20,14 +21,24 @@ class BranchActionForm(forms.Form):
...
     )
+    merge_strategy = forms.ChoiceField(
+        choices=BranchMergeStrategyChoices,
+        initial=BranchMergeStrategyChoices.ITERATIVE,
+        label=_('Merge Strategy'),
+        help_text=_('Strategy to use when merging changes.')
+    )
 
-    def __init__(self, branch, *args, allow_commit=True, **kwargs):
+    def __init__(self, branch, *args, allow_commit=True, action=None, **kwargs):
```
jeremystretch: With the introduction of the `merge_strategy`, my gut says now would be a good time to break this into separate forms for the sync, merge, and revert actions. We generally want to avoid conditionally deleting form fields because it can be confusing.

netbox_branching/templates/netbox_branching/branch_action.html
```diff
@@ -50,6 +50,9 @@
       <div class="row">
         <div class="col col-6 offset-3">
           {% render_field form.commit %}
+          {% if form.merge_strategy %}
+            {% render_field form.merge_strategy %}
+          {% endif %}
```
jeremystretch: Can we just use `{% render_form %}` here?

netbox_branching/merge_strategies/strategy.py
```diff
@@ -0,0 +1,64 @@
+"""
+Abstract base class for merge strategies.
+"""
```
jeremystretch: Redundant to class docstring?
```suggestion
```

netbox_branching/merge_strategies/strategy.py
```diff
@@ -0,0 +1,64 @@
...
+        MergeStrategy instance
+    """
+    from netbox_branching.choices import BranchMergeStrategyChoices
+    from .iterative import IterativeMergeStrategy
+    from .squash import SquashMergeStrategy
+
+    if strategy_name == BranchMergeStrategyChoices.SQUASH:
+        return SquashMergeStrategy()
+    else:
+        return IterativeMergeStrategy()
```
jeremystretch: Two suggestions:
1. Employ a dictionary to ease the introduction of potential future strategies.
2. Raise an exception if we encounter an unknown strategy name.
```suggestion
    try:
        return {
            BranchMergeStrategyChoices.SQUASH: SquashMergeStrategy(),
            BranchMergeStrategyChoices.ITERATIVE: IterativeMergeStrategy(),
        }[strategy_name]
    except KeyError:
        raise ValueError(f"Invalid strategy name: {strategy_name}")
```
We might also want to return only the class and have the caller handle instantiation, in case we decide to introduce arguments in the future.

netbox_branching/models/branches.py
```diff
@@ -537,20 +545,10 @@ def merge(self, user, commit=True):
...
-                models = set()
-
-                # Apply each change from the Branch
-                for change in changes:
-                    models.add(change.changed_object_type.model_class())
-                    with event_tracking(request):
-                        request.id = change.request_id
-                        request.user = change.user
-                        change.apply(self, using=DEFAULT_DB_ALIAS, logger=logger)
-                if not commit:
```
jeremystretch, line 554: I would prefer to keep the commit/abort logic contained to the branch itself, unless there's a specific reason to have the merge strategy be aware. This will avoid duplication of logic within each strategy, and ensure consistent commit behavior regardless of the selected strategy.

netbox_branching/models/branches.py
```diff
@@ -537,20 +545,10 @@ def merge(self, user, commit=True):
...
-                    raise AbortTransaction()
-
-                # Perform cleanup tasks
-                self._cleanup(models)
```
jeremystretch, line 558: I think it makes sense to perform post-merge cleanup as part of the merge strategy, because it's reasonable to assume that different strategies might have additional cleanup requirements. To that end, we should probably move the logic from `Branch._clean()` to the base `MergeStrategy` class.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+        """
+        Apply changes after collapsing them by object and ordering by dependencies.
+        """
+        models = set()
+
+        logger.info("Collapsing ObjectChanges by object...")
+        collapsed_changes = {}
+
+        for change in changes:
+            key = (change.changed_object_type.id, change.changed_object_id)
```
jeremystretch: Suggest utilizing the model's app & name for the key rather than the ContentType ID as it's much more human friendly when debugging. (I.e. `dcim.site:42` is more informative than `63:42`.) It also avoids complications that might arise from altered ContentType IDs.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+                model_class = change.changed_object_type.model_class()
+                collapsed = CollapsedChange(key, model_class)
+                collapsed_changes[key] = collapsed
+                logger.debug(f"New object: {model_class.__name__}:{change.changed_object_id}")
+
+            collapsed_changes[key].changes.append(change)
+
+        # Collapse each object's changes
+        for key, collapsed in collapsed_changes.items():
+            collapsed.collapse(logger)
```
jeremystretch: One thing to bear in mind is that we might be processing a very large number of changes (potentially tens of thousands). I'm curious whether the logic within `collapse()` can be adapted to apply iteratively for each change record processed, rather than storing them all first and then calling `collapse()` for each set of changes. This would enable us to avoid storing _all_ the changes in memory, as we would only have at most a single progressively-collapsed change record for each object at any time.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+
+            last_change = collapsed.last_change
+
+            logger.info(f"  [{i}/{len(ordered_changes)}] {collapsed.final_action.upper()} "
+                       f"{model_class.__name__}:{collapsed.key[1]} "
+                       f"(from {len(collapsed.changes)} original changes)")
+
+            with event_tracking(request):
+                request.id = last_change.request_id
+                request.user = last_change.user
```
jeremystretch: Because we're potentially compressing multiple changes that may come from different users, I'm not sure it makes sense to attribute them all to the user who made the most recent change. I would argue that wherever two or more changes are being squashed, the resulting change should be attributed to the user who is merging the branch, because that is the only person we should assume has full contextual awareness.

This topic probably deserves further discussion outside the context of this PR; I just wanted to flag it here for further consideration.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+
+class CollapsedChange:
+    """
+    Represents a collapsed set of ObjectChanges for a single object.
+    """
+    def __init__(self, key, model_class):
+        self.key = key  # (content_type_id, object_id)
+        self.model_class = model_class
+        self.changes = []  # List of ObjectChange instances, ordered by time
+        self.final_action = None  # 'create', 'update', 'delete', or 'skip'
```
jeremystretch: Might consider defining an Enum for these.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+        """
+        if not self.changes:
+            self.final_action = None
+            self.prechange_data = None
+            self.postchange_data = None
+            self.last_change = None
+            return
+
+        # Sort by time (oldest first)
+        self.changes = sorted(self.changes, key=lambda c: c.time)
```
jeremystretch: `changes` must already be in chronological order, no? In fact, we should raise an error if we attempt to append a new change record with a timestamp that comes before a record in the list as a sanity check.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+
+        has_delete = any(c.action == 'delete' for c in self.changes)
+        has_create = any(c.action == 'create' for c in self.changes)
+
+        logger.debug(f"  Collapsing {len(self.changes)} changes...")
+
+        if has_delete:
```
jeremystretch: Minor nit but I would break these into separate conditions (`if has_create and has_delete` and `if has_delete`) just to minimize indentation.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+                self.last_change = delete_change
+                return
+
+        # No DELETE - handle CREATE or UPDATEs
+        first_action = self.changes[0].action
+        first_change = self.changes[0]
+        last_change = self.changes[-1]
+
+        # Created (with possible updates) -> single create
+        if first_action == 'create':
```
jeremystretch: I think we can just reference `has_create` here, as it's not possible to have any other action prior to a create. (This might be another good opportunity for a sanity check.)

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+    """
+    Represents a collapsed set of ObjectChanges for a single object.
+    """
+    def __init__(self, key, model_class):
+        self.key = key  # (content_type_id, object_id)
+        self.model_class = model_class
+        self.changes = []  # List of ObjectChange instances, ordered by time
+        self.final_action = None  # 'create', 'update', 'delete', or 'skip'
+        self.prechange_data = None
+        self.postchange_data = None
```
jeremystretch: Can we initialize these to empty dicts to avoid having to do it under `collapse()`?
```suggestion
        self.prechange_data = {}
        self.postchange_data = {}
```

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+
+        for change in self.changes:
+            if change.postchange_data:
+                self.postchange_data.update(change.postchange_data)
+
+        logger.debug(f"  -> Action: UPDATE (collapsed {len(self.changes)} changes)")
+        self.final_action = 'update'
+        self.last_change = last_change
+
+    def apply(self, branch, using=DEFAULT_DB_ALIAS, logger=None):
```
jeremystretch: We're replicating a ton of logic from the ObjectChange class here under `apply()` and `undo()`. Instead of maintaining copies of this code (which are sure to drift over time), why don't we instead focus on returning a new ObjectChange instance with the necessary tweaks to its pre- and post-change data, and calling its `apply()`/`undo()` methods directly?

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+                    ct_field = getattr(instance, field.ct_field)
+                    fk_field = getattr(instance, field.fk_field)
+                    if ct_field and fk_field:
+                        setattr(instance, field.name, ct_field.get_object_for_this_type(pk=fk_field))
+
+            instance.full_clean()
+            instance.save(using=using)
+
+
+def _diff_object_change_data(action, prechange_data, postchange_data):
```
jeremystretch: This is essentially a copy of `ObjectChange.diff()`. Let's try to use that method directly rather than reproducing it.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+            f"Ordering by references exceeded maximum iterations ({max_iterations}). "
+            f"{len(remaining)} changes could not be ordered, possibly due to a complex cycle. "
+            f"Check the logs above for details."
+        )
+
+    logger.info(f"  Ordering by references completed: {len(ordered)} changes ordered")
+    return ordered
+
+
+def _order_collapsed_changes(collapsed_changes, logger):
```
jeremystretch: Should this be a method on SquashMergeStrategy?

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+    - UPDATEs that remove FK references happen before their associated DELETEs
+    - CREATEs happen before UPDATEs/CREATEs that reference them
+
+    Returns: ordered list of CollapsedChange objects
+    """
+    logger.info(f"Ordering {len(collapsed_changes)} collapsed changes...")
+
+    # Remove skipped objects
+    to_process = {k: v for k, v in collapsed_changes.items() if v.final_action != 'skip'}
+    skipped = [v for v in collapsed_changes.values() if v.final_action == 'skip']
```
jeremystretch: We can combine these into a single `for` loop.

netbox_branching/merge_strategies/squash.py
```diff
@@ -0,0 +1,790 @@
...
+    logger.info(f"  {len(skipped)} changes will be skipped (created and deleted in branch)")
+    logger.info(f"  {len(to_process)} changes to process")
+
+    if not to_process:
+        return []
+
+    # Reset dependencies
+    for collapsed in to_process.values():
+        collapsed.depends_on = set()
+        collapsed.depended_by = set()
```
jeremystretch: These are already set under `CollapsedChange.__init__()`.
```suggestion
```

netbox_branching/models/branches.py
arthanson: Just switching it back to the default. It is only saved state if you do a successful merge so the revert knows how to revert, otherwise it switches back to the default.

netbox_branching/collapse_merge.py
arthanson: oops, yeah

netbox_branching/forms/misc.py
arthanson: Done.

netbox_branching/models/branches.py
arthanson: It is split out because iterative merge strategy works against ObjectChange objects, but the squash works against the in-memory collapsed_change so the logic is different.  While it is probably possible to make the collapsed_change look identical to ObjectChange the amount of code common would still just be a few lines and I can imagine other merge strategies would possibly need different logic here.

netbox_branching/choices.py
arthanson: used pgettext_lazy

netbox_branching/models/branches.py
arthanson: It only needs to be set when merge takes place to know what strategy to use for revert - it is always default to iterative unless a different strategy is used to merge, then reset if it is reverted.

jnovinger:
@arthanson, looks like you have a CI failure due to some lingering merge conflict markers.

netbox_branching/models/branches.py
jeremystretch: It seems like the field should be left null then until the branch has actually been merged, since we don't choose a strategy until the branch is merged.

netbox_branching/models/branches.py
jeremystretch: I'm not sure I follow. I'm referring to the intentional raising of AbortRequest once the branch merge has been completed, to effect a dry run. This should be the same regardless of the merge strategy employed, because it's just rolling back the entire transaction at that point.

netbox_branching/views.py
```diff
@@ -238,13 +239,19 @@ def do_action(self, branch, request, form):
 @register_model_view(Branch, 'merge')
 class BranchMergeView(BaseBranchActionView):
     action = 'merge'
+    form = forms.BranchMergeForm
 
     def do_action(self, branch, request, form):
+        # Save the merge_strategy setting to the branch
+        branch.merge_strategy = form.cleaned_data.get('merge_strategy', 'iterative')
```
jeremystretch: We should rely on the form to control the default value.
```suggestion
        branch.merge_strategy = form.cleaned_data.get('merge_strategy')
```

netbox_branching/merge_strategies/squash.py
arthanson: refactored from collapse of entire list to a progressive add_change so each ObjectChange is added iteratively.  There is still a final list of all the collapsed changes, that is unavoidable - but the intermediate ObjectChages are not loaded all into memory.

One required change was the revert was passing in the changes in reverse order which didn't work for doing this way as you need to start at the beginning.  So added a revert_changes_ordering.

jnovinger:
FWIW, this looks good. Just holding off on approval/merging until I have `feature` cleaned up and can re-point this PR at it.

PR #372, opened 2025-11-27
Fixes #371: Ignore module import errors when discovering object migrators

### LINKED ISSUE #371: `ModuleNotFoundError` exception raised when attempting to merge a branch after upgrading to NetBox v4.4
### Plugin Version

v0.7.2

### NetBox Version

v4.4.7

### Python Version

3.12

### Steps to Reproduce

1. Start with NetBox v4.3.7 and netbox-branching v0.6.2
2. Create a branch
3. Upgrade to NetBox v4.4.7 and netbox-branching v0.7.2
4. Create a site in the branch (bug is not specific to the type of change being made)
5. Attempt to merge the branch

### Expected Behavior

The branch should merge successfully.

### Observed Behavior

The merge job fails with the following error message:

```
Error: ModuleNotFoundError("No module named 'thumbnail'")
```

After some investigation, I've discovered the root cause lies inside the `migrators()` method on the Branch model:

https://github.com/netboxlabs/netbox-branching/blob/9ce04cc693688281b8655b3c815e787ed0458a1e/netbox_branching/models/branches.py#L314-L320

Line 317 attempts to resolve the app to which a migration belongs. This fails on the `thumbnail.0001_initial` migration, because sorl-thumbnail installs as `sorl.thumbnail` but its migration is prefixed only with `thumbnail`.

As this method merely looks for object-specific migrators, it's probably safe to simply ignore any import failures.

jeremystretch:
### Fixes: #371

This specifically addresses a known issue with `sorl-thumbnail`, but applies a general workaround.

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -314,7 +314,12 @@ class Branch(JobsMixin, PrimaryModel):
         migrators = defaultdict(list)
         for migration in self.applied_migrations:
             app_label, name = migration.split('.')
-            module = importlib.import_module(f'{app_label}.migrations.{name}')
+            try:
+                module = importlib.import_module(f'{app_label}.migrations.{name}')
+            except ModuleNotFoundError:
+                logger = logging.getLogger('netbox_branching.branch')
+                logger.warning(f"Failed to load module for migration {migration}; skipping.")
+                continue
             for object_type, migrator in getattr(module, 'objectchange_migrators', {}).items():
                 migrators[object_type].append(migrator)
         return migrators


aoforb:
Thanks for this, I've applied the change manually and its fixed the issue.

netbox_branching/models/branches.py
```diff
@@ -314,7 +314,12 @@ def migrators(self):
         migrators = defaultdict(list)
         for migration in self.applied_migrations:
             app_label, name = migration.split('.')
-            module = importlib.import_module(f'{app_label}.migrations.{name}')
+            try:
+                module = importlib.import_module(f'{app_label}.migrations.{name}')
+            except ModuleNotFoundError:
+                logger = logging.getLogger('netbox_branching.branch')
```
bctiemann, line 320: I see the pattern in this file is to instantiate a new `logger` for every log line, which I guess makes sense since they're all routed differently for different purposes; but I'd think a more common/less chatty pattern would be to declare all the various loggers at the top instead of ad-hoc. Just a note for consideration though, not a reason to block this merge.

jeremystretch: I think in this case it's okay because we're only invoking the logger in the event a failure occurs, which should be rare. We _should_ probably devise a more robust logging strategy generally, but that's outside the scope of this bug fix.

Issue #384, opened 2025-12-15
Keep existing DATABASE["OPTIONS"] settings


seifert:
### Plugin Version

v0.7.3

### Proposed functionality

`DynamicSchemaDict` replaces `search_path` for branch databases. However, all existing keys except `options` are lost. For example, if the `sslmode` key is used, it is passed to `psycopg2.connect` when `default` database is requested, but not when branch database is requested.

I have prepared a simple [patch](https://github.com/netboxlabs/netbox-branching/compare/main...seifert:netbox-branching:dynamic-schema-dict-keep-options) that solves this problem.

### Use case

All options for branches databases will be kept, only `"options": "-c search_path=..." will be changed.

jnovinger:
After thinking about this for a bit, I believe this is actually a bug.

Thanks for volunteering, @seifert. Assigning to you now.

PR #385, opened 2025-12-16
Keep database options for branches connections

seifert:
### Fixes: #384

`DynamicSchemaDict` replaces `DATABASES["default"]["OPTIONS"]` by `{"options": f"-c search_path=..."}` for branch databases, so all existing keys are lost. For example, if the `sslmode` key is used, it is passed to `psycopg2.connect` when default database is requested, but not when branch database is requested. The bug is in the `DynamicSchemaDict.__getitem__` method.

### DIFF
--- a/netbox_branching/utilities.py
+++ b/netbox_branching/utilities.py
@@ -73,9 +73,9 @@ class DynamicSchemaDict(dict):
                 default_config = super().__getitem__('default')
                 return {
                     **default_config,
-                    "OPTIONS": {
-                        **default_config.get("OPTIONS", {}),
-                        "options": f"-c search_path={schema},{self.main_schema}"
+                    'OPTIONS': {
+                        **default_config.get('OPTIONS', {}),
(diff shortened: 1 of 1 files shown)


netbox_branching/utilities.py
```diff
@@ -71,11 +71,15 @@ def __getitem__(self, item):
                 track_branch_connection(item)
 
                 default_config = super().__getitem__('default')
+                options = {
+                    k: v
+                    for k, v in default_config.get("OPTIONS", {}).items()
+                    if k != "options"
+                }
+                options["options"] = f"-c search_path={schema},{self.main_schema}"
```
jnovinger:
```suggestion
                options = {
                	**default_config.get("OPTIONS", {}),
                	"options": f"-c search_path={schema},{self.main_schema}"
               	}
```

I think this can be simplified to something like this, which removes the need for a comprehension/for loop.

Issue #403, opened 2026-01-05
Update squash merge docs to reflect full scenario


jnovinger:
### Change Type

Addition

### Proposed Changes

#401 added documentation for the new squash merge strategy added in v0.8.0. However, the two examples miss a key point that makes the feature hard to grasp without.

In each case, a scenario is described where some object (a site `site_a`) somehow ends up being created outside of a particular working branch (either created directly in `main` or merged to `main` from some other branch) while also being created in the new branch.

Currently, the docs seem to imply that the squash merge strategy can handle this scenario on its own. That is incorrect. What it can do is allow branching users to "rescue themselves" by changing the conflicting object in their branch, then proceeding with a squash merge (an iterative merge will still fail in this case). Previously, such an edit wouldn't resolve a conflict, but due to the squashing/collapsing of edits in the squash strategy, the merge no longer runs in to a conflict.

PR #411, opened 2026-01-15
Fix squash merge documentation

mrmrcoleman:
Fix: https://github.com/netboxlabs/netbox-branching/issues/403

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -1,7 +1,7 @@
 # NetBox Branching Best Practices
 
 !!! note
-    This guidance applies to NetBox Branching version 0.7.3 and previous versions.
+    This document was last updated for NetBox Branching version 0.8.0.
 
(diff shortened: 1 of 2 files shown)


docs/best-practices.md
```diff
@@ -63,27 +62,11 @@ Consider the following scenario, using Site as an example object:
...
-2. Branch 2 is created.
-
-3. On Branch 1, a site is created with the name "Site A" and slug "sitea".
-
-4. On Branch 2, a site is also created with the name "Site A" and slug "sitea".
-
-5. Branch 1 is merged successfully.
-
-**Result:**
+**Recovery:**
```
jnovinger, line 67: This style of markup is not terribly consistent with our current practices. Normally, I would use another sub-header level, probably something like `#### Recovery` here.

However I do realize that very similar markup was already present in this file. In the interest of getting this docs fix out, let's ignore this for now and fix it later.

PR #414, opened 2026-01-21
307 exempt api/status from branching middleware

arthanson:

### DIFF
--- a/netbox_branching/utilities.py
+++ b/netbox_branching/utilities.py
@@ -15,7 +15,7 @@ from django.urls import reverse
 
 from netbox.plugins import get_plugin_config
 from netbox.utils import register_request_processor
-from .constants import BRANCH_HEADER, COOKIE_NAME, EXEMPT_MODELS, INCLUDE_MODELS, QUERY_PARAM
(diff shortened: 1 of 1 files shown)


netbox_branching/middleware.py
```diff
@@ -10,12 +10,20 @@
 
 
 class BranchMiddleware:
+    # Paths that should bypass branch activation
+    EXEMPT_PATHS = (
+        '/api/status/',
```
jeremystretch: We can't hard-code URLs: They need to be resolved by view name e.g. in case someone has a custom `BASE_PATH` defined.

netbox_branching/middleware.py
```diff
@@ -10,12 +11,23 @@
...
+    # View names that should bypass branch activation
+    EXEMPT_VIEW_NAMES = (
+        'api-status',
+    )
 
     def __init__(self, get_response):
         self.get_response = get_response
 
+        # Convert view names to paths during initialization
+        self.exempt_paths = tuple(reverse(name) for name in self.EXEMPT_VIEW_NAMES)
```
jeremystretch: Can we move `reverse()` to the definition of `EXEMPT_VIEW_NAMES`, to avoid having to call it for each request?

Issue #5, opened 2024-07-19
Automatically redirect to avoid 404s when activating or deactivating a branch


jeremystretch:
### Plugin Version

v0.2.0

### Proposed functionality

An example:

1. Create & activate a new branch
2. Create a new site
3. View the site
4. Deactivate the branch via the widget at top right

This returns a 404, because the request targets the site, which does not exist outside the branch.

Whenever the middleware detects a branch being activated/deactivated in the UI, it should check the response code. If it's a 404, redirect the user to the dashboard.

### Use case

This would help avoid confusing disruptions in the user's workflow.

PR #415, opened 2026-01-21
Automatically redirect to avoid 404s when activating or deactivating a branch

arthanson:
### Fixes: #5 

Whenever the middleware detects a branch being activated/deactivated in the UI, it checks the response code. If it's a 404, it redirects the user to the dashboard.

### DIFF
--- a/netbox_branching/tests/test_views.py
+++ b/netbox_branching/tests/test_views.py
@@ -148,12 +148,11 @@ class BranchMiddlewareTestCase(TransactionTestCase):
 
         # Follow the redirect and check for the warning message
         response = self.client.get(f'{site_url}?{QUERY_PARAM}={branch.schema_id}', follow=True)
-        messages = list(get_messages(response.wsgi_request))
-        self.assertEqual(len(messages), 2)  # One success message for activation, one warning
-        # Check that the warning message mentions the branch name
-        warning_messages = [m for m in messages if 'does not exist' in str(m)]
-        self.assertEqual(len(warning_messages), 1)
+        messages_list = list(get_messages(response.wsgi_request))
+        warning_messages = [m for m in messages_list if 'does not exist' in str(m)]
+        self.assertGreaterEqual(len(warning_messages), 1, "Expected at least one warning message")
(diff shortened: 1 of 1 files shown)


netbox_branching/middleware.py
```diff
@@ -30,9 +31,17 @@ def __call__(self, request):
...
+
             if branch:
                 response.set_cookie(COOKIE_NAME, branch.schema_id)
-            elif QUERY_PARAM in request.GET:
+            elif branch_change:
                 response.delete_cookie(COOKIE_NAME)
 
+            # Redirect to dashboard if branch activation/deactivation results in 404
+            if branch_change and response.status_code == 404:
+                messages.warning(request, "The requested object does not exist in the current branch.")
```
jnovinger: Can we provide a more explicit message to the user, ideally mentioning what branch and some information about the requested item?

netbox_branching/middleware.py
```diff
@@ -30,9 +31,17 @@ def __call__(self, request):
...
+                return HttpResponseRedirect('/')
```
jnovinger: I'm curious about the chose order of operations here.

Since we're doing this new check and redirect _after_ the cookie set/delete check and that set/delete is being done on the original response, it seems like we could end up in a case where we think the cookie is being deleted, but actually isn't. Is that intentional?

netbox_branching/tests/test_views.py
```diff
@@ -50,3 +58,60 @@ def setUpTestData(cls):
...
+        self.client.force_login(self.user)
+
+    def tearDown(self):
+        # Manually tear down any dynamic connections created for branches
+        for branch in Branch.objects.all():
+            if branch.connection_name in connections:
+                connections[branch.connection_name].close()
+
+    @override_settings(LOGIN_REQUIRED=False)
+    def test_redirect_on_404_during_branch_deactivation(self):
```
jnovinger, line 80: Any chance we could also get a test that exercises this in the other direction? That is when the new middleware behavior is triggered by activating a branch instead of deactivating?

jnovinger (approved):

PR #427, opened 2026-02-11
Fix #422 Allow deletion of objects that have been deleted in main

### LINKED ISSUE #422: Delete within a branch should not show warning if item has been deleted in main
### Plugin Version

v0.8.1

### NetBox Version

v4.5.1

### Python Version

v3.12

### Steps to Reproduce

1. create a site in main 's1'
2. create and activate a branch 'b1'
3. switch back to main branch
4. delete the site 's1'
5. switch to branch 'b1'
6. delete the site 's1' in the branch

### Expected Behavior

The site should be allowed to be deleted as this allows squash merge to fix issues.

However this would also cause an issue with the iterative merge strategy if it was deleted, so a potential fix would need to be made there as well.

### Observed Behavior

Get Message "Cannot delete site 's1' because it has been deleted in the main branch. Sync with the main branch to update." and Cannot delete the site.

arthanson:
### Fixes: #422 

* Removed warning message if deleting an object in the branch and already deleted in main.
* Added test case for this scenario
* also test case for merge to make sure squash merge can merge branch where object has edits, was deleted in main then deleted in branch (main reason for this fix).
* cleaned up the merge tests made more DRY

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -52,6 +52,24 @@ The maximum total number of branches that can exist simultaneously, including me
 
 ---
 
+## `job_timeout`
+
+Default: `3600` (1 hour)
+
+The maximum time in seconds that long-running branch operations (sync, merge, revert) are allowed to execute before timing out. This timeout applies to background jobs that process large branches.
+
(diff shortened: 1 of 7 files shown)


jnovinger:
@arthanson , looks like this has conflicts with `main`.

jnovinger (approved):

Issue #440, opened 2026-02-19
Add tests for API endpoints for sync, merge and revert endpoints


arthanson:
### Plugin Version

v0.8.1

### Proposed functionality

Add tests to test_api that covers the sync, merge and revert API functionality.

### Use case

These tests are currently missing from the tests cases and should be covered.

### External dependencies

N/A

PR #441, opened 2026-02-19
#440 Add API tests for sync, merge, revert APIs

arthanson:
### Fixes: #440 

Extends the API tests cases for sync, merge and revert which were missing.

### DIFF
--- a/netbox_branching/tests/test_api.py
+++ b/netbox_branching/tests/test_api.py
@@ -28,6 +28,7 @@ class BaseAPITestCase:
         ContentType.objects.get_for_model(Branch)
 
     # TODO: Remove when dropping support for NetBox v4.4
+    @staticmethod
     def create_token(user):
         try:
             # NetBox >= 4.5


netbox_branching/tests/test_api.py
```diff
@@ -6,48 +6,42 @@
...
 
-# TODO: Remove when dropping support for NetBox v4.4
-def create_token(user):
-    try:
-        # NetBox >= 4.5
-        from users.choices import TokenVersionChoices
-        token = Token(version=TokenVersionChoices.V1, user=user)
-        token.save()
-    except ImportError:
-        # NetBox < 4.5
```
jeremystretch, line 24: We need to keep this fallback logic in place as netbox-branching v0.8 still supports NetBox v4.4.

netbox_branching/tests/test_api.py
```diff
@@ -194,3 +174,177 @@ def test_patch_status_archived_blocked(self):
 
         branch.refresh_from_db()
         self.assertEqual(branch.status, BranchStatusChoices.MERGED)
+
+
+class BranchSyncAPITestCase(BaseAPITestCase, TransactionTestCase):
+
+    def test_sync_endpoint_success(self):
```
jeremystretch, line 192: Can we extend this and its peer methods in the other test cases to check for the creation of a job to perform the sync/merge/revert?

netbox_branching/tests/test_api.py
```diff
@@ -13,41 +14,41 @@
...
+    def create_token(self, user):
+        try:
+            # NetBox >= 4.5
+            from users.choices import TokenVersionChoices
+            token = Token(version=TokenVersionChoices.V1, user=user)
+        except ImportError:
+            # NetBox < 4.5
+            token = Token(user=user)
+        token.save()
+        return token.token
```
jeremystretch: Did you test this on NetBox v4.4? IIRC the `token` attribute is exclusive to v4.5+. (Note that the original code returns `token.key` for v4.4.)

Issue #86, opened 2024-09-06
Omit "empty" values from change data for created & deleted objects


jeremystretch:
### Plugin Version

v0.4.0

### Proposed functionality

When displaying the serialized representation of objects that have been created or deleted within a branch, omit any empty values (null, empty strings, empty arrays, etc.) from the output. For example, creating a site named "New Site" will show the following data in the "after" state of the change:

```
asns: []
comments: 
custom_fields: {}
description: 
facility: 
group: None
latitude: None
longitude: None
name: New Site
physical_address: 
region: None
shipping_address: 
slug: new-site
status: active
tags: []
tenant: None
time_zone: None
```

Omitting empty values leaves only fields which contain meaningful data, namely the `name` and `slug` fields:

```
name: New Site
slug: new-site
```

This more closely resembles the behavior when modifying an existing object, where only the modified fields are shown.

### Use case

This would reduce noise in the change log and simplify human review.

PR #444, opened 2026-02-25
86 omit empty values from display data in branch changes

arthanson:
### Fixes: #86

old display:
Monosnap b2 | NetBox 2026-02-25 12-59-35

new display:
Monosnap b2 | NetBox 2026-02-25 12-59-08

### DIFF
--- a/netbox_branching/tables/tables.py
+++ b/netbox_branching/tables/tables.py
@@ -34,8 +34,8 @@ BEFORE_DIFF = """
 {% if record.action == 'create' %}
     {{ ''|placeholder }}
 {% elif record.action == 'delete' %}
-    <pre class="p-0">{% for k, v in record.diff.pre|compact_items %}{{ k }}: {{ v }}
-{% endfor %}</pre>
+    <pre class="p-0">{% for k, v in record.diff.pre.items %}{% if not v|is_empty %}{{ k }}: {{ v }}
+{% endif %}{% endfor %}</pre>
 {% else %}
     <pre class="p-0">{% for k, v in record.diff.pre.items %}{{ k }}: {{ v }}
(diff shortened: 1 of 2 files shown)


netbox_branching/templatetags/filters.py
```diff
@@ -0,0 +1,13 @@
+from django import template
+
+register = template.Library()
+
+EMPTY_VALUES = (None, '', [], {})
```
jeremystretch: We can reference `django_filters.constants.EMPTY_VALUES` instead of defining our own.

netbox_branching/templatetags/filters.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@register.filter
+def compact_items(value):
```
jeremystretch: Rather introducing a filter which forces a second iteration, how about a simple filter e.g. `is_empty` that can be applied to the value directly within the `for` loop in the template?

netbox_branching/tables/tables.py
```diff
@@ -30,17 +30,25 @@
 """
 
 BEFORE_DIFF = """
+{% load filters %}
 {% if record.action == 'create' %}
     {{ ''|placeholder }}
+{% elif record.action == 'delete' %}
+    <pre class="p-0">{% for k, v in record.diff.pre|compact_items %}{{ k }}: {{ v }}
+{% endfor %}</pre>
```
jeremystretch: With the proposed `is_empty` filter:
```suggestion
    <pre class="p-0">{% for k, v in record.diff.pre %}{% if not v|is_empty %}{{ k }}: {{ v }}
{% endif %}{% endfor %}</pre>
```

PR #470, opened 2026-03-18
#469 recalculate cable paths when squash merge

### LINKED ISSUE #469: Squash merge does not recalculate cable paths
### Plugin Version

v0.8.2

### NetBox Version

v4.5.5

### Python Version

3.13

### Steps to Reproduce

This was found when creating the new test suites for branching, currently the test test_merge_cable_path_recalculation works for iterative merge but is disabled for squash merge. You can replicate by removing the `@unittest.skip` at https://github.com/netboxlabs/netbox-branching/blob/main/netbox_branching/tests/test_iterative_merge.py#L969

### Expected Behavior

squash merge should work the same as iterative merge in maintaining cable paths.

### Observed Behavior

squash merge does not recalculate the cable paths.

arthanson:
### Fixes: #469

requires [#21880](https://github.com/netbox-community/netbox/pull/21880) - tests will fail until this is added in netbox.


jeremystretch (changes requested):
We can't implement any logic specific to cables in the netbox-branching plugin, as it would be extremely difficult (if not impossible) to maintain long-term. Instead, see if you can devise a way to trigger the tracing logic using a signal or something similar.

arthanson:
rewrote with a change for NetBox  [#21880](https://github.com/netbox-community/netbox/pull/21880) so it is generic.

arthanson:
@claude /review

Issue #425, opened 2026-02-11
Create a merge report for errored merge


arthanson:
### Plugin Version

v0.8.1

### Proposed functionality

Create a merge report that users can view after a failed merge that contains suggestions on next steps to resolve any merge issues.

Current idea is to attach a json object that when the merge encounters error conditions it can add text that with suggestions on resolution steps the user can take to get a mergeable branch.  We would then possibly need a view to parse the json and show a more user-friendly report or we could just show the json.

Examples:
* if constraint errors are hit it can direct them to use the squash merge strategy
* If an not-found error occurs because an object has been deleted in main then the user can be directed to sync the branch
* If we do decided to special case handle deleted in main objects (like re-creating them) then a note can be made about this.

It would also be possible to show some stats that might be helpful, such as:

branch: mybranch
Date merge ran: 00-00-00
merge strategy: iterative

Issues encountered:

constraint: object xxx could not be merged - you can rename the issue or use the squash merge strategy.

number objects created: xx
number objects updated: yy
number objects deleted: zz

### Use case

Currently when a merge fails all the guidance the user is given in the job log which lists the error, it is then up to the end-user to figure out how to correct the issues.  This would allow us to give guidance for at least common failure scenarios so it is easier for the user to resolve the issues.

### External dependencies

N/A

PR #477, opened 2026-03-23
425 Create a merge report

arthanson:
### Fixes: #425 

This pretty much just specifically handles reporting for unique constraints right now.  I was looking at a secondary scenario of removing an item you have a FK reference to - but that is actually spawned a new bug [#478](https://github.com/netboxlabs/netbox-branching/issues/478) - which will take longer to fix, once that is fixed it can be added to this merge report.  This is a start to get it out there, we will probably add to it in the future as we identify specific failure scenarios and their suggested workarounds.

in squash and iterative merge we annotated the exception info to save the model class and object-id so we know what object got the exception.  In job.py we also store the totals of the items changed and what merge strategy was used as data could change after the job was run.

We also only show the report from the last job that was run. Older jobs will have had data changes and are no longer applicable so there is no real reason to allow showing those, you can also get the log information from the job if need be.

When you run a merge or sync if there are errors, the report button will appear in the alert area for the failed merge/sync:
Monosnap b2 | NetBox 2026-03-23 11-05-07

There is also a tab to show the logs for help in debugging:

Monosnap b2 - Sync Report | NetBox 2026-03-23 11-04-42

Here is a merge report - note the guidance is different between sync and merge as for sync you can't tell them to run the squash merge to fix it.  Also - if there were pending sync when you ran the job it will report that with guidance that we recommend running sync first before doing a merge.

Monosnap b2 - Merge Report | NetBox 2026-03-23 11-06-19

**Note:** Unfortunately we can only capture one unique constraint issue at a time, so if you have 3 conflicting names the report will. only show one, which you can fix and re-run again which will show the second one.

Scenarios for recommendations:

### Unique Constraint
1. create a branch
2. create site "s1" in main
3. create site "s1" in branch
4. try to merge

**Note:** we need to parse the error message to get the field in `_analyze_integrity_error` if you have a non-english locale postgresql this will not be able to pull the field name out so in merge failure we will still show the guidance to rename the conflicting object, but we won't show the field name in the table. I don't think there is a good way around that as it is language dependent.

### DIFF
--- a/netbox_branching/error_report.py
+++ b/netbox_branching/error_report.py
@@ -60,9 +60,9 @@ def _get_field_from_constraint(table_name, constraint_name):
                 return constraint.fields[0] if len(constraint.fields) == 1 else None
         for field in model._meta.get_fields():
             col = getattr(field, 'column', None)
-            if col and getattr(field, 'unique', False):
-                if constraint_name in (f'{table_name}_{col}_key', f'{table_name}_{col}_uniq'):
-                    return field.name
+            auto_names = (f'{table_name}_{col}_key', f'{table_name}_{col}_uniq')
+            if col and getattr(field, 'unique', False) and constraint_name in auto_names:
+                return field.name
         break
     return None
 
--- a/netbox_branching/views.py
+++ b/netbox_branching/views.py
@@ -72,7 +72,11 @@ class BranchView(generic.ObjectView):
             'last_job': last_job,
             'last_job_errored': last_job is not None and last_job.status == JobStatusChoices.STATUS_ERRORED,
             'last_merge_job': last_merge_job,
-            'last_merge_job_errored': last_merge_job is not None and last_merge_job == last_job and last_merge_job.status == JobStatusChoices.STATUS_ERRORED,
+            'last_merge_job_errored': (
+                last_merge_job is not None
+                and last_merge_job == last_job
+                and last_merge_job.status == JobStatusChoices.STATUS_ERRORED
+            ),
             'conflicts_count': ChangeDiff.objects.filter(branch=instance, conflicts__isnull=False).count(),
         }
 


netbox_branching/templates/netbox_branching/branch_job_report.html
```diff
@@ -0,0 +1,256 @@
...
+                      {{ entry.object_str }}
+                    {% else %}
+                      —
+                    {% endif %}
+                  </td>
+                  <td>{{ entry.field|default:"—" }}</td>
+                  <td>{% if entry.value %}<code>{{ entry.value }}</code>{% else %}—{% endif %}</td>
+                  <td>{{ entry.message }}</td>
+                </tr>
+                <tr class="table-light">
```
jeremystretch: Suggest moving this outside of the table into a separate card to improve readability and separate advisory text from job data.

netbox_branching/templates/netbox_branching/branch_job_report.html
```diff
@@ -0,0 +1,256 @@
...
+
+{% block content %}
+  {# Report tab #}
+  <div class="tab-pane show active" id="report" role="tabpanel" aria-labelledby="report-tab">
+
+    {% if merge_strategy %}
+      <p class="text-secondary mb-3">
+        <strong>{% trans "Merge Strategy" %}:</strong> {{ merge_strategy|title }}
+      </p>
+    {% endif %}
```
jeremystretch: This should go inside a panel. Suggest combining this with the change statistics in a single data panel. Try to stay consistent with the rest of the NetBox UI.

netbox_branching/templates/netbox_branching/branch.html
```diff
@@ -33,13 +33,24 @@
...
-        <i class="mdi mdi-alert"></i>
-        The <a href="{{ last_job.get_absolute_url }}">last job</a> ({{ last_job.name }}) failed with an error:
-      </h4>
-      <div class="text-secondary">{{ last_job.error }}</div>
+  {% if last_job_errored %}
+    <div class="alert alert-danger d-flex justify-content-between align-items-start" role="alert">
+      <div>
+        <h4 class="alert-title">
+          <i class="mdi mdi-alert"></i>
+          The <a href="{{ last_job.get_absolute_url }}" class="alert-link">last job</a> ({{ last_job.name }}) failed with an error.
```
jeremystretch: Missing translation support (wrap with `{% trans %}`)

netbox_branching/templates/netbox_branching/branch.html
```diff
@@ -33,13 +33,24 @@
...
+        </h4>
+        {% if last_job.data.report %}
+          <div class="text-secondary">{{ last_job.data.report|length }} error{{ last_job.data.report|length|pluralize }} occurred.</div>
```
jeremystretch: Missing translation support (wrap with `{% trans %}`)

netbox_branching/error_report.py
```diff
@@ -0,0 +1,192 @@
...
+
+def annotate_validation_error(exc, model_class, object_id, content_type_id):
+    """Attach branch operation context to a ValidationError before re-raising."""
+    exc.netbox_branching_model = model_class
+    exc.netbox_branching_object_id = object_id
+    exc.netbox_branching_content_type_id = content_type_id
+
+
+# PostgreSQL error codes
+PG_UNIQUE_VIOLATION = '23505'
```
jeremystretch: Let's move this to `constants.py` and add a comment explaining its usage.

netbox_branching/jobs.py
```diff
@@ -19,13 +26,56 @@
...
+
+
 def get_job_log(job):
     """
     Initialize and return the job log.
     """
-    job.data = {
-        'log': []
-    }
+    job.data = {'log': []}
```
jeremystretch: Maybe out of scope for the PR, but this seems like it should only set `log` if it's not already defined in the job data.

netbox_branching/jobs.py
```diff
@@ -19,13 +26,56 @@
 )
 
 
+def _snapshot_changes_summary(changes_qs):
```
jeremystretch: Any reason for this not to live on MergeBranchJob?

netbox_branching/jobs.py
```diff
@@ -19,13 +26,56 @@
...
+    """
+    Compute a JSON-serializable changes summary from a queryset, snapshotted at job start time.
+    Stored as {'creates': {'app_label.model': count}, ...} so it remains accurate even if the
+    branch state changes later.
+    """
+    changes = changes_qs.values('action', 'changed_object_type_id', 'changed_object_id')
+
+    object_actions = defaultdict(set)
+    for change in changes:
```
jeremystretch: Might want to avoid loading the entire set of changes into memory, to avoid performance issues at scale. Another option is to query the aggregate counts for each type directly from the database.

netbox_branching/views.py
```diff
@@ -140,6 +144,54 @@ def get_children(self, request, parent):
...
+            for entry in last_job.data['report']:
+                object_url = None
+                object_str = None
+                ct_id = entry.get('content_type_id')
+                obj_id = entry.get('object_id')
+                if ct_id and obj_id:
+                    try:
+                        ct = ContentType.objects.get_for_id(ct_id)
+                        obj = ct.get_object_for_this_type(pk=obj_id)
+                        object_url = f'{obj.get_absolute_url()}?_branch={instance.schema_id}'
```
jeremystretch: Might want to check for the present of `get_absolute_url` on the object, just to be safe. (Thinking about models from other plugins.)

netbox_branching/views.py
```diff
@@ -140,6 +144,54 @@ def get_children(self, request, parent):
         return parent.get_unmerged_changes().order_by('time')
 
 
+@register_model_view(Branch, 'job-report')
+class BranchJobReportView(generic.ObjectView):
+    queryset = Branch.objects.all()
+    template_name = 'netbox_branching/branch_job_report.html'
+
+    def get_extra_context(self, request, instance):
```
jeremystretch, line 199: Just a suggestion, but I would try to break this monolithic block into at least one separate private method, as it can be difficult to follow.

netbox_branching/error_report.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def _table_to_model(table_name):
+    """Return the verbose_name for a Django model given its DB table name, or None."""
+    for model in apps.get_models():
```
jeremystretch, line 55: We should probably cache a dictionary mapping to avoid iterating through the entire set of models for each change.

netbox_branching/error_report.py
```diff
@@ -0,0 +1,192 @@
...
+
+def get_merge_recommendations(entry, merge_strategy=None):
+    """Compute actionable recommendations for a failed merge or revert operation."""
+    is_squash = merge_strategy == BranchMergeStrategyChoices.SQUASH
+
+    error_type = entry.get('type')
+    field = entry.get('field', '')
+    value = entry.get('value', '')
+
+    if error_type == 'unique_constraint':
```
jeremystretch, line 199: Just a suggestion, but I find it cleaner to define the recommendation messages as constants and reference them by name, e.g. ` return RENAME_IN_BRANCH.format(...)`. This help separate the user-facing text from the decision logic.

netbox_branching/views.py
```diff
@@ -64,6 +68,7 @@ def get_extra_context(self, request, instance):
             'stats': stats,
             'latest_change': latest_change,
             'last_job': last_job,
+            'last_job_errored': last_job is not None and last_job.status == JobStatusChoices.STATUS_ERRORED,
```
jeremystretch, line 73: This will catch a failed job of _any_ type, but BranchJobReportView specifically filters for merge jobs:

```python
last_job = instance.jobs.filter(name=MergeBranchJob.Meta.name).order_by('created').last()
```

So the "view report" button might be presented for a failed job (e.g. a sync failure) that doesn't have any report data. Not sure how much of a concern this is.

netbox_branching/error_report.py
```diff
@@ -0,0 +1,191 @@
...
+    if sqlstate == PG_UNIQUE_VIOLATION:
+        # diag attributes are locale-independent catalog values (psycopg3).
+        table_name = getattr(diag, 'table_name', None) if diag else None
+
+        # Format: "Key (field)=(value) already exists." — locale-dependent; returns None
+        # for non-English PostgreSQL locales, falling back to the generic message.
+        field = None
+        value = None
+        if diag and diag.message_detail:
+            detail_match = re.search(r'Key \((.+?)\)=\((.+?)\)', diag.message_detail)
```
jeremystretch, line 89: This seems shaky for non-English locales. Would `diag.column_name` be a suitable alternative?

arthanson: PostgreSQL only populates diag.column_name for NOT NULL violations — for unique violations it's always None. The locale-independent alternative is diag.constraint_name, which IS populated for unique violations.  The regex is used as a fallback if that constraint_name isn't set.

Also fixed a bug that was introduced in some previous changes where OBJECT and VALUE columns weren't showing all the time, added some fallbacks for those.

arthanson:
@jeremystretch had to make some larger code changes for the last comments so resubmitting for (hopefully) last review. See response to your comment above.

Issue #213, opened 2025-02-05
Display a warning when a branch is at risk of becoming stale


jeremystretch:
### Plugin Version

v0.5.3

### Proposed functionality

This happens when a branch has not been synchronized in some time, and its `last_sync` time is at risk of exceeding the configuration changelog retention time. We should display a warning for branches that are within a configurable amount of time of the cutoff.

### Use case

This will help draw attention to branches that have not been kept up to date.

### External dependencies

N/A

PR #492, opened 2026-03-30
213 Display warning when branch is becoming stale

arthanson:
### Fixes: #213 

Displays a warning message when branch is about to become stale:

Monosnap b1 | NetBox 2026-03-30 13-26-26

Here is a command line to set all branches to a past last_sync so the message will show up:

```
python netbox/manage.py shell -c "from datetime import timedelta; from django.utils import timezone; from netbox_branching.models import Branch;  print(Branch.objects.exclude(status__in=['merged','archived']).update(last_sync=timezone.now()-timedelta(days=85)), 'rows updated')"
```

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -1,5 +1,6 @@
 import importlib
 import logging
+import math
 import random
 import string
 from collections import defaultdict
@@ -298,7 +299,7 @@ class Branch(JobsMixin, PrimaryModel):
         if not (changelog_retention := get_config().CHANGELOG_RETENTION):
             return None
         stale_at = self.last_sync + timedelta(days=changelog_retention)
-        return (stale_at - timezone.now()).days
+        return math.ceil((stale_at - timezone.now()).total_seconds() / 86400)
 
     @property
     def is_stale(self):


netbox_branching/constants.py
```diff
@@ -9,9 +9,13 @@
     'INCLUDE_MODELS',
     'QUERY_PARAM',
     'SKIP_INDEXES',
+    'STALE_WARNING_THRESHOLD',
 )
 
 
+# Number of days before staleness at which to display a stale warning
+STALE_WARNING_THRESHOLD = 7
```
jeremystretch: I suggest making this a plugin configuration parameter as users will almost certainly want some control over the warning period.

netbox_branching/models/branches.py
```diff
@@ -300,6 +300,22 @@ def is_stale(self):
...
 
+    @property
+    def is_stale_warning(self):
+        """
+        Indicates whether the branch is approaching the staleness threshold and at risk of becoming stale.
+        """
+        if self.last_sync is None:
+            return False
+        if not (changelog_retention := get_config().CHANGELOG_RETENTION):
+            return False
```
jeremystretch: These lines are redundant to `is_stale()` above. Suggest implementing a private method (e.g. `_days_until_stale()`) to return `self.last_sync < timezone.now() - timedelta(days=changelog_retention)` (or `None` as appropriate) for use within both `is_stale()` and `is_stale_warning()`.

netbox_branching/models/branches.py
```diff
@@ -300,6 +300,22 @@ def is_stale(self):
             return False
         return self.last_sync < timezone.now() - timedelta(days=changelog_retention)
 
+    @property
+    def is_stale_warning(self):
```
jeremystretch: Suggest renaming this as "is stale warning" sounds odd.
```suggestion
    def stale_warning(self):
```

netbox_branching/templates/netbox_branching/branch.html
```diff
@@ -42,6 +42,15 @@ <h4 class="alert-title">
       <div class="text-secondary">{{ last_job.error }}</div>
     </div>
   {% endif %}
+  {% if object.is_stale_warning %}
+    <div class="alert alert-warning d-flex align-items-center" role="alert">
+      <i class="mdi mdi-alert fs-1 text-warning me-3"></i>
+      <div>
+        <h4 class="alert-title text-warning mb-0">{% trans "Branch is at risk of becoming stale" %}</h4>
+        {% blocktrans %}This branch has not been synced recently and is at risk of becoming stale. Sync the branch to avoid losing the ability to merge.{% endblocktrans %}
```
jeremystretch: It would be helpful to cite the time remaining until the branch becomes stale.

Issue #461, opened 2026-03-10
Provisoning a branch fails due to a model using mutli-table inheritance


PaulLocksley:
### Plugin Version

0.8.1

### NetBox Version

4.5.3

### Python Version

3.12

### Steps to Reproduce

1. Create a model that uses Mutli-table inheritance in your database
2. Attempt to provision a branch

I can create a minimal plugin to reproduce if you would like exact steps to reproduce, Just wanted to see if this was even a bug first or just an expected .

### Expected Behavior

I would hope the branch was provisioned correctly. 

I'm not sure if this is the expected behavior, I did some searching but couldn't find any discussion on this so wanted to see, I know netbox branching doesn't intend to support every possible netbox model. 

### Observed Behavior

Provisioning a new branch fails with: 
[2026-03-10 01:05:06,392] DEBUG Creating table branch_585y1usm.otherplugin_l2e
[2026-03-10 01:05:06,567] ERROR column "id" of relation "otherplugin_l2e" does not exist

I think it's failing here:
                    # Get the name of the sequence used for object ID allocations (if one exists)
                    cursor.execute(
                        "SELECT pg_get_serial_sequence(%s, 'id')", [table]
                    )
as it's the same error I get when running the query manually. 

This model uses [multi-table-inheritance]( https://docs.djangoproject.com/en/6.0/topics/db/models/#multi-table-inheritance) for its ID field so the id actually comes from a separate table.

PR #495, opened 2026-03-31
#461 document that multi-table inheritance isn't supported

arthanson:
### Fixes: #461 

Documents that multi-table inheritance is not supported in branching.

### DIFF
--- a/docs/plugin-development.md
+++ b/docs/plugin-development.md
@@ -30,7 +30,7 @@ Models that do **not** use `ChangeLoggingMixin` are ineligible for branching sup
 
 These models are still accessible from within a branch, but changes to them made inside a branch are **not isolated** — they affect the main schema immediately, just as if no branch were active.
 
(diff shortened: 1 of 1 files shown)


docs/plugin-development.md
```diff
@@ -28,6 +28,8 @@ Models that do **not** use `ChangeLoggingMixin` are ineligible for branching sup
 - Junction/through tables managed entirely by many-to-many fields
 - Models you explicitly register as exempt (see below)
 
+**Multi-table inheritance is not supported.** Models that use Django's [multi-table inheritance](https://docs.djangoproject.com/en/6.0/topics/db/models/#multi-table-inheritance) are not compatible with NetBox Branching. Each model in a branch must map to a single, self-contained table.
```
jnovinger: The MTI paragraph lands right before "These models are still accessible from within a branch, but changes to them made inside a branch are **not isolated**..." which was written for the no-`ChangeLoggingMixin` models in the bullet list above. Reading it linearly it now sounds like it applies to MTI too. MTI models are actually worse: provisioning fails outright rather than silently leaking changes to main.

I'd move the MTI paragraph after that "still accessible" paragraph to break the implied connection, and add a sentence about the actual failure mode. Something like: "Attempting to provision a branch when such models are present will result in a provisioning error."

PR #502, opened 2026-04-06
478 fix merging references to objects that were deleted in main

### LINKED ISSUE #478: Reference to object that gets deleted in main causes merge to fail
### Plugin Version

v0.8.2

### NetBox Version

v4.5.5

### Python Version

3.12

### Steps to Reproduce

1. create site s1
2. in branch create location l1 with FK to s1 (site)
3. in main delete s1
4. sync - this is successful
5. merge (iterative or squash) fail

The issue is there is still a changes-ahead for the creation of l1 and that is causing both merges to fail as the reference in it to site: s1 which was deleted makes it fail.

This is an edge-case where the object in the branch references an object in main that is deleted

### Expected Behavior

the merge should work after the sync.

### Observed Behavior

The merge fails.

arthanson:
### Fixes: #478 

Fixes a bug where merging a branch would crash if an object created in the branch was cascade-deleted from the branch schema during a prior sync (e.g. its FK parent was deleted in main). In this state, the branch's ObjectChange log still contains the original CREATE (and any subsequent UPDATEs), causing the merge to attempt replaying changes for an object that never made it to main.

The fix addresses this at the source: during sync, a temporary pre_delete handler identifies any branch-originated objects that are cascade-deleted as a side effect of applying a DELETE change, and writes a synthetic DELETE ObjectChange for each one. This keeps the changelog accurate and lets both merge strategies handle the case naturally - squash collapses CREATE + DELETE into a no-op, and iterative skips changes for objects that can no longer be created or updated in main.

**Note:** if you don't sync in this scenario before doing a merge an error will still occur, but this is by design as we recommend users to sync before merging and it is more complicated and separate code to try to detect and fix this during merge.

Iterative merge will still fail as it normally does with items deleted in main. Squash merge will handle this.


arthanson:
@claude /review

jeremystretch (commented):
I'm not sure this is the optimal approach to solve this. It seems the root issue is that we're not catching and recording the cascading deletion of the child location when the branch is synced from main.

When you delete a site in main, for example, a deletion changelog record is recorded for each deleted location. This behavior should ideally be the same within the branch: When the site is deleted from the branch, a new deletion record for the child location should be created in the branch's changelog. At least, that's what I would expect to happen. If the deletion record exists, squashing the branch should work.

How feasible is to generate deletion records for branch-originated objects deleted during a sync?

arthanson:
@jeremystretch changed to create a delete change-object.

arthanson:
@claude /review

arthanson:
re claude review - fixed item 1. Others not needed.

netbox_branching/models/branches.py
```diff
@@ -460,6 +464,73 @@ def can_archive(self):
...
+            ObjectChange.objects.using(self.connection_name).create(
+                action=ObjectChangeActionChoices.ACTION_DELETE,
+                changed_object_type=ct,
+                changed_object_id=obj_pk,
+                object_repr=obj_repr,
+                prechange_data=prechange_data,
+                postchange_data=None,
+                user=user,
+                user_name=user.username if user else '',
+                request_id=uuid.uuid4(),
```
jeremystretch: It may be confusing to pass a unique UUID for each change record resulting from the sync, because they will appear to come from unrelated changes. There's no objectively correct answer here, but IMO it would be helpful to use the same UUID for all records: This will at least allow correlating them.

If you want to take this approach, it probably makes sense to stage a dummy request under `sync()` like we do for `merge()` and `revert()` and pass through its request ID:

```python
request = RequestFactory().get(reverse('home'))
```

jeremystretch (approved):
Nice work! Had one suggestion piggybacking on Claude's feedback but deferring to your judgment. Approving as-is; happy to follow up if you want to make the tweak.

netbox_branching/models/branches.py
arthanson: Changed to pass a common UUID. the RequestFactory wasn't working here.

jeremystretch (approved):

PR #508, opened 2026-04-14
#506 - Set SAMESITE for branch cookie

### LINKED ISSUE #506: Branch cookie not cleared when branch becomes non-ready; missing SameSite attribute causes branch writes to leak to main
### Plugin Version

v0.8.3

### NetBox Version

v4.5.8

### Python Version

3.13

### Steps to Reproduce

There is not an easy way to reproduce this locally just by the browser..

1. Create a branch via the UI
2. Activate it in the browser (cookie is now set)
3. Force the branch to a non-ready status via the Django shell:

```
  python netbox/manage.py shell
  from netbox_branching.models import Branch
  from netbox_branching.choices import BranchStatusChoices
  b = Branch.objects.get(name='your-branch-name')
  b.status = BranchStatusChoices.ARCHIVED
  b.save(provision=False)
```

4. In the browser - Navigate anywhere in the UI — the branch cookie is still present in the browser but the branch is no longer ready
5. Create an object (e.g. a region) — it will be created in main, not the branch
6. Check that no error or warning is shown, and the cookie was not cleared in the response headers

You can confirm the cookie persists by checking Application > Cookies in browser devtools and watching the active_branch cookie value never get cleared across requests.

### Expected Behavior

When a branch is active, all writes should be routed to that branch's schema. The active_branch cookie should be set with SameSite, Secure, Domain, and Path attributes consistent with Django's SESSION_COOKIE_* settings, so it is treated reliably by all browsers and proxy configurations

It should also be automatically cleared when the referenced branch is no longer in READY status, with the user returned to main context.

### Observed Behavior

The active_branch cookie is set with no SameSite, Secure, Domain, or Path attributes. In browsers or configurations that do not default to SameSite=Lax, the cookie may be dropped or ignored, causing all writes to route silently to main while the user believes they are working in a branch.

arthanson:
### Fixes: #506 

fixes an issue where the branch cookie was being set without a SameSite attribute. This matters because browsers that don't default to Lax would silently drop the cookie on cross-site navigation, causing branch writes to fall through to the main schema. The fix mirrors Django's SESSION_COOKIE_* settings when setting or deleting the branch cookie, so it behaves consistently with the rest of the session.

The PR also handles stale cookies: if a cookie references a branch that's no longer active (e.g. it was archived or merged since the cookie was set), the middleware now detects this and clears the cookie automatically.

### DIFF
--- a/netbox_branching/middleware.py
+++ b/netbox_branching/middleware.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
 from django.http import HttpResponseBadRequest, HttpResponseRedirect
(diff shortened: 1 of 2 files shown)


jeremystretch:
@claude /review This is a pretty simple PR. Just note any changes that might have unexpected effects.

jeremystretch (approved):

PR #515, opened 2026-04-17
513 changediff

### LINKED ISSUE #513: ChangeDiff diff properties crash with AttributeError when original or modified is None
### Plugin Version

0.9.0

### NetBox Version

4.5.3

### Python Version

3.12

### Steps to Reproduce

1. Create a branch
2. Apply assurance data to it (this creates objects on the branch via diode, which produces ChangeDiff records with `original=None` for CREATE actions)
3. Go to Change Management → select your branch → Changes tab

### Expected Behavior

The Changes tab should render without errors, showing the created objects.

### Observed Behavior

Server error:

```text
AttributeError: 'NoneType' object has no attribute 'items'
```

### Root Cause

`ChangeDiff.original` and `ChangeDiff.modified` are nullable `JSONField`s. The signal receiver in `signal_receivers.py:207` correctly sets `original=instance.prechange_data_clean or None` — which is `None` for CREATE actions (no prior state).

However, three `cached_property` methods on `ChangeDiff` call `.items()` on these fields without checking for `None`:

* `altered_in_modified` (line \~235): `self.modified.items()` and `k in self.original`
* `original_diff` (line \~275): `self.original.items()`
* `modified_diff` (line \~285): `self.modified.items()`

Note that `altered_in_current` and `current_diff` already have proper `None` guards — the fix is applying the same pattern to the three missing properties.

### Related Issues

* [#250](<https://github.com/netboxlabs/netbox-branching/issues/250>) — same error class, partial fix in `_update_conflicts` only
* [NPL-751](https://linear.app/netboxlabs/issue/NPL-751/attributeerror-in-changediff-update-conflicts-when-selfmodified-is) — same error in `_update_conflicts` with `self.modified`, closed as unreproducible
* [#512](<https://github.com/netboxlabs/netbox-branching/issues/512>) -- manually created, didn't trigger label workflows, this issue replaces it

### Linked

[NBE-862](https://linear.app/netboxlabs/issue/NBE-862/bug-or-change-management-change-log-for-branch-with-assurance-data-is)

arthanson:
### Fixes: #513 

The diff field was silently dropped from the API response for any ChangeDiff record with a CREATE or DELETE action. For CREATE records original is None, and for DELETE records modified is None. The three cached properties `altered_in_modified`, `original_diff`, and `modified_diff` called .items() on those fields without null guards, raising AttributeError. 

Because Python's attribute lookup treats AttributeError from a descriptor as "attribute not found", DRF silently omitted diff from the serialized output rather than returning a 500. The fix adds null guards matching the pattern already used by altered_in_current and current_diff.

Tests are added specifically for the ChangeDiff diff function making sure its returns are correct and API test cases are added that show this issue. Note: `test_changediff_list_update_action` does not show this error but is included for completeness for a full CRUD test for returned data.

### DIFF
--- a/netbox_branching/models/changes.py
+++ b/netbox_branching/models/changes.py
@@ -232,6 +232,8 @@ class ChangeDiff(models.Model):
         """
         Return the set of attributes altered in the branch schema.
         """
+        if self.original is None or self.modified is None:
+            return set()
         return {
             k for k, v in self.modified.items()
             if k in self.original and v != self.original[k]
@@ -272,6 +274,8 @@ class ChangeDiff(models.Model):
         """
         Return a key-value mapping of all attributes in the original state which have been modified.
         """
+        if self.original is None:
+            return {}
         return {
             k: v for k, v in self.original.items()
             if k in self.altered_fields
@@ -282,6 +286,8 @@ class ChangeDiff(models.Model):
         """
         Return a key-value mapping of all attributes which have been modified within the branch.
         """
+        if self.modified is None:
+            return {}
         return {
             k: v for k, v in self.modified.items()
             if k in self.altered_fields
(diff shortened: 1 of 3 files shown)


jeremystretch:
@claude /review

jeremystretch (approved):

PR #516, opened 2026-04-17
#494 allow merge branch if file missing from S3 storage

### LINKED ISSUE #494: Branch cannot merge if using S3 storage and file not found
### Plugin Version

v0.8.3

### NetBox Version

v4.4.5

### Python Version

3.13

### Steps to Reproduce

1. setup to use S3 storage
2. create and activate a branch 'b1'
3. go to a site and add an image attachment and choose "Create and Add Another"
4. add the same image again.
5. go to merge the branch

Note this will create a duplicated image name in the ObjectChange due to NetBox issue: [#21801](https://github.com/netbox-community/netbox/issues/21801). 

The issue is at: https://github.com/netboxlabs/netbox-branching/blob/main/netbox_branching/models/changes.py#L58

```
        if self.action == ObjectChangeActionChoices.ACTION_CREATE:
            if hasattr(model, 'deserialize_object'):
                ...
            except (FileNotFoundError) as e:
                # If a file was deleted later in this branch it will fail here
                # so we need to ignore it. We can assume the NetBox state is valid.
                logger.warning(f'Ignoring missing file: {e}')
            instance.save(using=using)
```
However S3 storage gives a different error than FileNotFound. - specifically

```
        "Applying change 298402 using default (Extras | image attachment Install created by user(@site)",
        "An error occurred (403) when calling the HeadObject operation: Forbidden"
```
The except should catch both errors from local file storage and the S3 Storages specific one.

### Expected Behavior

The branch should merge successfully.

### Observed Behavior

The branch cannot be merged.

arthanson:
### Fixes: #494 

Note that S3 needs to be configured to return 403 instead of 404 for unknown file, which is a security practice called bucket enumeration prevention.

This means a genuine misconfigured IAM policy (where the credentials truly lack access to the bucket) would also be silently swallowed and logged as "Ignoring missing file."  But given the context - this code runs during a merge, where the files were already successfully written during normal branch operation - a 403 on HeadObject at merge time is almost certainly the S3 enumeration-protection behavior, not a credentials problem. A real credential failure would have surfaced much earlier when the file was first saved.

### DIFF
--- a/netbox_branching/utilities.py
+++ b/netbox_branching/utilities.py
@@ -17,6 +17,7 @@ from netbox.utils import register_request_processor
 
 from .constants import BRANCH_HEADER, COOKIE_NAME, EXEMPT_MODELS, EXEMPT_PATHS, INCLUDE_MODELS, QUERY_PARAM
 from .contextvars import active_branch
+from .models.changes import _FILE_NOT_FOUND_EXCEPTIONS
 
 # Thread-local storage for tracking branch connection aliases (matches Django's approach)
 # Note: Aliases are tracked once and never removed, matching Django's pattern where
@@ -250,7 +251,7 @@ def update_object(instance, data, using):
 
     try:
         instance.full_clean()
-    except (FileNotFoundError) as e:
+    except _FILE_NOT_FOUND_EXCEPTIONS as e:
         # If a file was deleted later in this branch it will fail here
         # so we need to ignore it. We can assume the NetBox state is valid.
         logger.warning(f'Ignoring missing file: {e}')
--- a/netbox_branching/utilities.py
+++ b/netbox_branching/utilities.py
@@ -15,9 +15,16 @@ from django.urls import reverse
 from netbox.plugins import get_plugin_config
 from netbox.utils import register_request_processor
 
-from .constants import BRANCH_HEADER, COOKIE_NAME, EXEMPT_MODELS, EXEMPT_PATHS, INCLUDE_MODELS, QUERY_PARAM
+from .constants import (
+    _FILE_NOT_FOUND_EXCEPTIONS,
+    BRANCH_HEADER,
+    COOKIE_NAME,
+    EXEMPT_MODELS,
+    EXEMPT_PATHS,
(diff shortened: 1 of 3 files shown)


arthanson:
@claude /review

arthanson:
Claude review - not major concerns.

jnovinger (changes requested):
The fix, as is, looks great. However, `update_object()` in `utilities.py:253` has the same `FileNotFoundError`-only catch around `full_clean()`. That path gets hit for UPDATE actions during both `apply()` and `undo()`, so a modified image attachment on a branch would hit the same S3 403 failure. Since `_FILE_NOT_FOUND_EXCEPTIONS` is already defined at module level in `changes.py`, you could import it in `utilities.py` to keep both sites consistent.

Also worth noting (not for this PR): `undo()` for ACTION_DELETE calls `full_clean()` at `changes.py:126` with no file-missing handling at all. Could be a future gap if someone deletes an object with file attachments and then reverts the branch.

arthanson:
@jnovinger added your suggestion for utilities.py but had to move _FILE_NOT_FOUND_EXCEPTIONS into constants to deal with circular imports.

netbox_branching/utilities.py
```diff
@@ -250,7 +258,7 @@ def update_object(instance, data, using):
 
     try:
         instance.full_clean()
-    except (FileNotFoundError) as e:
+    except _FILE_NOT_FOUND_EXCEPTIONS as e:
         # If a file was deleted later in this branch it will fail here
         # so we need to ignore it. We can assume the NetBox state is valid.
         logger.warning(f'Ignoring missing file: {e}')
```
jnovinger: I notice that we're re-raising some exceptions (e.g. 400-403) in `changes.py:61` (or so). Should we be doing that here as well?

arthanson:
@claude /review

arthanson:
@jnovinger Good suggestion. Made it into a common function called by both as I hate it not being DRY.

jnovinger (approved):

PR #520, opened 2026-04-22
Fixes #519: Disable Merge button on branches with no changes

### LINKED ISSUE #519: Disable Merge button on branches with no changes
### Plugin Version

0.9.0

### Proposed functionality

The Merge button on the branch detail view (`buttons/branch_merge.html`) currently only checks permissions. It should also check whether the branch has any unmerged changes via `Branch.get_unmerged_changes().exists()`. If no changes exist, render a disabled button with a tooltip explaining why (e.g., "No changes to merge").

### Use case

Prevents users from attempting to merge a branch that has no changes, which is a confusing no-op. Improves the signal about what actions are actually available on a given branch.

### External dependencies

None

jnovinger:
### Fixes: #519

Check `branch.get_unmerged_changes().exists()` before enabling the Merge button. When no changes exist, render a disabled button with a Bootstrap tooltip explaining why ("No changes to merge" or "Insufficient permissions"). Uses the same span-wrapper tooltip pattern as NetBox core's `datasource.html`.

### Testing

- [ ] Branch detail: branch with unmerged changes shows active Merge button
- [ ] Branch detail: branch with no changes shows disabled Merge button with "No changes to merge" tooltip
- [ ] Branch detail: user without merge permission sees disabled Merge button with "Insufficient permissions" tooltip
- [ ] CR detail (approved): branch with changes shows active Merge button
- [ ] CR detail (approved): branch with no changes shows disabled Merge button with "No changes to merge" tooltip
- [ ] CR detail (not approved): Merge button not rendered (existing `is_approved` gate, unchanged)

### DIFF
--- a/netbox_branching/templates/netbox_branching/buttons/branch_merge.html
+++ b/netbox_branching/templates/netbox_branching/buttons/branch_merge.html
@@ -1,10 +1,12 @@
 {% load i18n %}
-{% if perms.netbox_branching.merge_branch %}
+{% if perms.netbox_branching.merge_branch and branch.get_unmerged_changes.exists %}
   <a href="{% url 'plugins:netbox_branching:branch_merge' pk=branch.pk %}" class="btn btn-primary">
     <i class="mdi mdi-merge"></i> {% trans "Merge" %}
   </a>
 {% else %}
-  <button type="button" class="btn btn-primary disabled" aria-disabled="true">
-    <i class="mdi mdi-merge"></i> {% trans "Merge" %}
-  </button>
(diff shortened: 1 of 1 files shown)


jnovinger:

jnovinger:

The original pattern was copied from NetBox core's `datasource.html`, which also uses the non-Bootstrap `inline-block` class. NetBox doesn't define a custom `.inline-block` CSS rule, so it was effectively a no-op. Fixed in f6571a3 to use Bootstrap's `d-inline-block` utility class.

arthanson (approved):

PR #524, opened 2026-04-27
Fixes #522: Share button toast, typo, and i18n for activation messages

### LINKED ISSUE #522: Fix share button toast, typo, and i18n for activation messages
### Plugin Version

0.9.0

### NetBox Version

4.5.7

### Python Version

3.13

### Steps to Reproduce

**Share button toast:**
1. Activate a branch
2. Click the share button on any object detail page
3. Observe a "Activated branch X" toast appears, even though the branch was already active

**Share button typo:**
1. Hover over the share button
2. Observe the tooltip reads "Link to the this page" (extra "the")

**Activation messages not i18n-wrapped:**
1. Review `utilities.py:299-303`
2. Observe activation/deactivation messages use f-strings instead of `gettext()`

### Expected Behavior

1. Clicking share should copy the URL to clipboard (or at minimum not show a redundant activation toast when the branch is already active)
2. Tooltip should read "Link to this page"
3. All user-facing messages should be wrapped in `gettext()` for i18n

### Observed Behavior

1. Redundant "Activated branch X" toast fires on every share button click
2. Typo in share button tooltip
3. Activation/deactivation messages are not translatable

_Split from a combined issue that also covered netbox-changes._

jnovinger:
### Fixes: #522

- Suppress the redundant "Activated branch" toast when clicking the share button for an already-active branch (compares query param against cookie)
- Fix "Link to the this page" typo in share button tooltip
- Wrap all activation/deactivation messages in `gettext()` for i18n (utilities.py and middleware.py)
- Add test for the toast suppression behavior

### DIFF
--- a/netbox_branching/middleware.py
+++ b/netbox_branching/middleware.py
@@ -2,6 +2,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
 from django.http import HttpResponseBadRequest, HttpResponseRedirect
+from django.utils.translation import gettext as _
 
 from .constants import COOKIE_NAME, EXEMPT_PATHS, QUERY_PARAM
 from .utilities import get_active_branch, is_api_request
@@ -72,7 +73,9 @@ class BranchMiddleware:
                 requested_url = request.path
                 messages.warning(
                     request,
-                    f"The requested object at {requested_url} does not exist in {branch_name}."
+                    _("The requested object at {url} does not exist in {branch_name}.").format(
+                        url=requested_url, branch_name=branch_name
+                    )
                 )
 
                 # Create redirect response and apply cookie operations to it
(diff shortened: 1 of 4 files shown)


jnovinger:

jnovinger:

jnovinger:
> **Potential double-message on genuine branch activation**

Addressed in d859d4f. Added a `_branch_activation_notified` request attribute flag to `get_active_branch()` so the toast only fires on the first call per request. Same deduplication pattern used in netbox-changes for `_warned_unsatisfiable_rules`. Added message count assertion to `test_activate_branch` to catch regressions.

Intentionally kept both `get_active_branch()` call sites (BranchMiddleware and ActiveBranchContextManager) rather than having the context manager read `request.active_branch`. The dual-call is by design: middleware handles cookie management while the context manager handles the ContextVar for database routing, and ExitStack's LIFO cleanup ordering depends on both running independently. See #414 (exempt path deduplication) and the review discussion on #500 (comment by @arthanson explaining why `request.active_branch` was added as a separate attribute rather than replacing the ContextVar call).

arthanson (approved):

PR #532, opened 2026-05-04
#529 allow specifying migrations to skip

### LINKED ISSUE #529: _fake_for_branch does not handle RunPython and AlterModelOptions operations
### Plugin Version

v0.9.1

### NetBox Version

4.5.x

### Python Version

3.12

### Steps to Reproduce

1. Install netbox-branching alongside a plugin that has non-branchable models (e.g. netbox-changes, where all models are exempt)
2. Create and provision a branch
3. Upgrade the plugin to a version that includes both a schema migration (AddField) and a subsequent data migration (RunPython) targeting the same non-branchable model
4. Run `branch.migrate()` on the existing branch

For a concrete example using netbox-changes:
- Migration 0009 adds `reviewed_through_change_id` to `Review` (a non-branchable model)
- Migration 0010 is a `RunPython` that backfills that field

### Expected Behavior

Both migrations should be faked on the branch schema, since they exclusively affect non-branchable models. The branch should reach `READY` status without errors.

### Observed Behavior

Migration 0009 (`AddField`) is correctly faked because the operation has a `model_name` attribute and the model is non-branchable. Migration 0010 (`RunPython`) is NOT faked and actually executes, because `RunPython` operations have no `model_name` attribute.

`_fake_for_branch` (`models/branches.py:61-93`) skips operations where `getattr(operation, 'model_name', None)` is `None`, so `has_model_operations` stays `False` and the function returns `False` (don't fake). The data migration then fails because it references a field/column that was never created in the branch schema.

The same gap affects `AlterModelOptions`, which uses a `name` attribute instead of `model_name`. Migration 0012 in netbox-changes (`AlterModelOptions`) is also incorrectly run instead of faked.

Verified locally by calling `_fake_for_branch` against each migration:

```
=== netbox_changes ===
  0009_review_reviewed_through_change_id: fake=True   ops=['AddField']
  0010_backfill_reviewed_through_change_id: fake=False ops=['RunPython']
  0012_policy_custom_permissions: fake=False           ops=['AlterModelOptions']
```

arthanson:
### Fixes: #529 

Adds a param `fake_on_branch` developers can add to migrations to tell it to skip when doing the branching migrations.  Has no effect otherwise.

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -60,15 +60,29 @@ __all__ = (
 
 def _fake_for_branch(migration):
     """
-    Return True if all model-specific operations in a migration affect only non-branchable
-    models. Such migrations should be faked on branch schemas to prevent RunSQL operations
-    from inadvertently acting on the main (public) schema via the search_path.
+    Return True if a migration should be faked when applied to a branch schema, False otherwise.
 
-    Migrations with no model-specific operations (e.g. pure RunSQL) are not faked, as we
-    cannot determine their intent without executing them.
+    Decision order:
+    1. If the migration module sets a ``fake_on_branch`` attribute, that value is respected
+       directly: ``True`` forces faking, ``False`` forces the migration to run.
+    2. Otherwise, fall back to a heuristic: fake migrations whose model-specific operations
+       affect only non-branchable models. This prevents RunSQL operations from inadvertently
+       acting on the main (public) schema via the search_path.
+
+    Migrations with no model-specific operations (e.g. pure RunSQL or RunPython) are not faked
+    by the heuristic, as we cannot determine their intent without executing them. Authors of
+    such migrations should set ``fake_on_branch`` explicitly when needed.
 
     SeparateDatabaseAndState operations are not supported and will be skipped with an error.
     """
+    # Check for an explicit per-migration override
+    try:
+        module = importlib.import_module(f'{migration.app_label}.migrations.{migration.name}')
+    except ModuleNotFoundError:
+        module = None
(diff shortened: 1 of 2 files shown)


jnovinger:

netbox_branching/models/branches.py
jnovinger, line 94: My local Claude points out that there are some schema operations (e.g. CreateModel, DeleteModel) that don't have a `model_name` instance attribute and instead have a `name` attribute. Should we be checking for that as well to cover this potential gap?

jnovinger (commented):
Two suggestions:

1. The existing heuristic checks `model_name` but several model-targeting operations (`AlterModelOptions`, `CreateModel`, `DeleteModel`, `RenameModel`) use `name` instead. That's the gap that caused the `AlterModelOptions` failure in the issue. The `fake_on_branch` escape hatch covers it, but the heuristic could also handle these automatically with something like:

```python
model_name = getattr(operation, 'model_name', None) or getattr(operation, 'name', None)
```

Was this considered? The flag is still valuable for `RunPython`/`RunSQL` either way.

2. `_fake_for_branch` has no test coverage. Given the new override behavior, tests for the three states (`True`, `False`, absent) would be worthwhile.

netbox_branching/models/branches.py
arthanson: I think that is outside the scope of this PR. If we did make those changes I don't think it wouldn't affect this issue at all. The failure was in 0010 migration which is just RunSQL.  Doing this change would require more testing, etc.. So potentially valid for a future issue, but I don't think we should do that mixed in with this.

jnovinger (approved):

PR #534, opened 2026-05-04
#533 Fix check_object_accessible_in_branch 

### LINKED ISSUE #533: Fix check_object_accessible_in_branch to perform a true existence check
### Plugin Version

v1.0.0

### Proposed functionality

Replace `model.objects.get(pk=…)` with `model.objects.filter(pk=…).exists()` in `check_object_accessible_in_branch` (`netbox_branching/signal_receivers.py`). The function is a boolean predicate; .exists() compiles to SELECT 1 … WHERE id = %s LIMIT 1 rather than selecting every column on the model. Return values are identical for every caller.

### Use case

The current SELECT * raises ProgrammingError whenever the in-memory model class disagrees with the active schema's column list - for example, a column renamed in one schema but not the other, rolling deploys where the model is newer than the table, or mid-migration windows. When that happens, the post-save change-tracking pipeline aborts even though the question being asked (does the row exist?) doesn't depend on any of those columns. A pk-only existence check is immune because the primary key can't be renamed, so the boolean answer is correct regardless of column drift.

Secondary benefits: the existence check is cheaper at the wire (one sentinel value instead of every column on the model), avoids materializing the row in PostgreSQL, and replaces try/except DoesNotExist exception flow with a plain boolean expression. 

### External dependencies

N/A

arthanson:
### Fixes: #533 

Change to use and exist() check instead of a try, get, except.  This is cheaper operation and doesn't load all the model fields which was causing issues in the revert in branching for custom objects as the schema was different between the states of the object as they are dynamically created and only one is in memory at once.

### DIFF
--- a/netbox_branching/signal_receivers.py
+++ b/netbox_branching/signal_receivers.py
@@ -55,13 +55,8 @@ def check_object_accessible_in_branch(branch, model, object_id):
     if f'{model._meta.app_label}.{model._meta.model_name}' in INCLUDE_MODELS:
         return True
 
-    # Check whether the object exists in main
     with deactivate_branch():
-        try:
-            model.objects.get(pk=object_id)
-        except model.DoesNotExist:
-            pass
-        else:
+        if model.objects.filter(pk=object_id).exists():
             return True
 
     # Object doesn't exist in main - check if it was created in the branch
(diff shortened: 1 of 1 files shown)


arthanson:
@claude /review

jeremystretch (approved):

PR #535, opened 2026-05-05
#531 - fix branch sync for MPTT nested records

### LINKED ISSUE #531: Branch Sync may fail even when no (conflicting) changes present in branch
### Plugin Version

0.9.1

### NetBox Version

4.4.10

### Python Version

3.12.3

### Steps to Reproduce

1. Create new Platform `Windows Server OS`
2. Create a new branch, *do not activate*
3. Update Platform `Windows Server OS` name to `Windows Server`
4. Create new Platform `Windows`
5. Update Platform `Windows Server` to have parent `Windows`
6. Attempt to sync branch

### Expected Behavior

Branch is synced successfully. Platform changes are replicated properly.

### Observed Behavior

The job fails with the error `Cannot assign self or child platform as parent.`:

```
{
    "log": [
        "Syncing branch Test (branch_ia873u7z)",
        "Found 3 changes to sync",
        "Setting branch status to syncing",
        "Applying change 287387 using schema_branch_ia873u7z (DCIM | platform Windows Server updated by user)",
        "Updating platform Windows Server OS",
        "Applying change 287388 using schema_branch_ia873u7z (DCIM | platform Windows created by user)",
        "Applying change 287389 using schema_branch_ia873u7z (DCIM | platform Windows Server updated by user)",
        "Updating platform Windows Server",
        "{'parent': ['Cannot assign self or child platform as parent.']}"
    ]
}
```

arthanson:
### Fixes: netboxlabs/netbox-branching#531

Adds a function to clear the MPTT fields.  It is not great doing this as it ties it directly to MPTT but MPTT currently does not provide a direct way to do this and instead has a rebuild but that would force the entire tree to be rebuilt which would be a large performance hit.  Centralized the clearing into one function that would be easy to replace in the future.

NestedGroupModel (the MPTT base for Platform/Region/etc.) sets order_insertion_by = ('name',). When a new root is inserted on main, MPTT keeps roots ordered alphabetically and renumbers tree_id on existing roots in-place via raw SQL to make room. Those side-effect updates don't go through save(), so they generate no ObjectChange records.

In the issue's repro:

* Main starts with Windows Server OS (tree_id=1).
* Renamed to Windows Server (tree_id=1).
* New root Windows inserted. Because "Windows" < "Windows Server" alphabetically, MPTT gives Windows tree_id=1 and silently bumps Windows Server to tree_id=2. The bump is invisible to the changelog.
* The CREATE Windows ObjectChange's postchange_data records tree_id=1.

When the branch syncs:

* Windows Server OS is renamed → Windows Server (still tree_id=1 in branch).
* CREATE Windows is applied via deserialize_object + DeserializedObject.save(), which calls Model.save_base(raw=True) and bypasses MPTT entirely. Branch now has two roots with tree_id=1.
* Reparenting Windows Server → Windows triggers MPTT's \_move_root_node, which sees node.tree_id == target.tree_id (==1) and raises InvalidMove.

## The fix

In netbox_branching/models/changes.py, the CREATE branch of ObjectChange.apply() now detects MPTT models, clears their tree fields, and calls instance.object.save() (regular path) instead of the raw DeserializedObject.save() — letting MPTT compute tree_id/lft/rght/level against the branch's local state. M2M data is then applied manually since we bypassed DeserializedObject.save().

rebuild() would fix the collision, but it walks every row of the model's table to renumber all trees from parent_id — O(n) per affected MPTT model per sync, regardless of how many rows actually changed. It also reassigns tree_id by its own ordering, which can diverge from main's tree_id values for unrelated roots. Clearing fields on just the  CREATE and letting MPTT compute against local state is bounded to the changed rows and leaves untouched trees alone.


arthanson:
@claude /review

arthanson:
with respect to Claude review, #1 is a non issue.

jnovinger (approved):
Two suggestions, both non-blocking:

- The test asserts `name` and `parent_id` but doesn't verify MPTT tree consistency (`level`, `lft < rght`, ancestor chain). A quick `self.assertEqual(ws_in_branch.level, 1)` or `get_ancestors()` check would add confidence. CI passes and MPTT would blow up on subsequent operations if the tree were inconsistent, so this is minor.

- The `undo()` MPTT handling (concern #2 above) could be a follow-up issue if undo of deleted MPTT objects is a supported workflow.

The root cause analysis in the PR description is thorough and the `clear_mptt_fields()` utility is well-factored. Approving.

PR #525, opened 2026-04-29
Fixes #447: Fix IntegrityError when creating devices with front/rear port mappings in a branch

arthanson:
### Fixes: #447 

NetBox 4.5's PortMapping table lacks ChangeLoggingMixin, so BranchAwareRouter sends its writes to the main schema. When a device is created inside a branch, the resulting PortMapping.bulk_create() fails with a FK violation because the device only exists in the branch schema.

Fix is a one-line addition to INCLUDE_MODELS, matching the existing pattern for TaggedItem and CablePath. Also adds test_related_models.py covering all applicable INCLUDE_MODELS entries.

Issue #447, opened 2026-03-02
Error when creating non-racked device in branch


PaulR282:
### Plugin Version

0.8.1

### NetBox Version

4.5.3

### Python Version

3.12

### Steps to Reproduce

1. Create a non-racked device

### Expected Behavior

Non-racked device should be created

### Observed Behavior

```
Traceback (most recent call last):
  File "/opt/netbox/netbox/netbox/middleware.py", line 128, in __call__
    username = request.META[self.header]
               ^^^^^^^^^^^^^^^^^^^^^^^^^

During handling of the above exception ('HTTP_REMOTE_USER'), another exception occurred:
  File "/opt/netbox-4.5.3/venv/lib/python3.12/site-packages/django/db/backends/base/base.py", line 303, in _commit
    return self.connection.commit()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox-4.5.3/venv/lib/python3.12/site-packages/psycopg/connection.py", line 304, in commit
    self.wait(self._commit_gen())
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox-4.5.3/venv/lib/python3.12/site-packages/psycopg/connection.py", line 483, in wait
    return waiting.wait(gen, self.pgconn.socket, interval=interval)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "psycopg_c/_psycopg/waiting.pyx", line 241, in psycopg_c._psycopg.wait_c
    <source code not available>
  File "/opt/netbox-4.5.3/venv/lib/python3.12/site-packages/psycopg/_connection_base.py", line 585, in _commit_gen
    yield from self._exec_command(b"COMMIT")
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox-4.5.3/venv/lib/python3.12/site-packages/psycopg/_connection_base.py", line 483, in _exec_command
    raise e.error_from_result(result, encoding=self.pgconn._encoding)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

The above exception (insert or update on table "dcim_portmapping" violates foreign key constraint "dcim_portmapping_device_id_eb86e378_fk_dcim_device_id"
[...]

PR #540, opened 2026-05-06
539 Fix branch migrations

### LINKED ISSUE #539: in-Branch migrations fail
### Plugin Version

v1.0.0

### NetBox Version

v4.6.0

### Python Version

3.13

### Steps to Reproduce

1. Install 4.4.10, Branching 0.8.3, Changes 0.4.3 (NBE 2.0.1 in effect)
2. Create 1 site + 1 device type (eg: EX4200 from the demo data) + 1 device role
3. Create a Branch
4. In that branch, create 1 site, and 2 devices on that site 
5. Update to NetBox 4.5.x, Branching 1.0.0 or 1.0.1, changes 0.4.3
6. Try to run migrations on the branch

### Expected Behavior

migrations work

### Observed Behavior

Migrations Fail with a programming error

```
{
    "log": [
        "Migrating branch test (branch_s3xy608w)",
        "Setting branch status to migrating",
        "Faking migration users.0013_user_remove_is_staff (no branchable models affected)",
        "Faking migration users.0014_users_token_v2 (no branchable models affected)",
        "Applying migration users.0015_owner",
        "Applying migration extras.0134_owner",
        "Applying migration dcim.0217_poweroutlettemplate_color",
        "Applying migration dcim.0218_owner",
        "Applying migration dcim.0219_devicetype_device_count",
        "Applying migration dcim.0220_cable_profile",
        "Applying migration dcim.0221_cable_connector_positions",
        "Applying migration dcim.0222_port_mappings",
        "column dcim_frontporttemplate.rear_port_position does not exist\nLINE 1: ...\".\"description\", \"dcim_frontporttemplate\".\"type\", \"dcim_fron...\n                                                             ^"
    ]
}
```

arthanson:
### Fixes: #539 

Fixes branch upgrade migration from NetBox 4.4 to 4.6, where data migrations in dcim.0222_port_mappings were failing with column dcim_frontporttemplate.rear_port_position does not exist.

## Changes

**1. Branch.migrate() - activate the branch during migration (models/branches.py)**

Wrap the migration loop in with activate_branch(self): so the active-branch contextvar is set while RunPython data migrations execute. Without this, the router has no signal to route queries away from main, even when supports_branching() would otherwise return True.

**2. Branch.migrate() - mark failed branches as FAILED (models/branches.py)**

On exception during migration, set status to BranchStatusChoices.FAILED instead of READY. Previously a partially-migrated branch was left activatable which is incorrect.

**3. supports_branching() - resolve historical models to live registry (utilities.py)**

RunPython data migrations receive historical models from the migration's StateApps. Those don't inherit ChangeLoggingMixin from the live Python class, so the old issubclass() check always returned False, and the BranchAwareRouter sent branch-aware queries to main. Resolve the model via apps.get_model(app_label, model_name) against the live registry before the class-hierarchy check, with a LookupError fallback for models that no longer exist live.

**4. Add dcim.porttemplatemapping to INCLUDE_MODELS (constants.py)**

The DeviceType-level port template mapping table (added in NetBox 4.6) needs branch-aware routing for the same reason dcim.portmapping does (PR #525, issue #447). Without this, the data migration that populates it fails with an FK violation against dcim_devicetype in main.

 **5. Isolate RunSQL operations from main during branch migration (models/branches.py)**

During Branch.migrate(), wrap each RunSQL.database_forwards call so search_path is restricted to the branch schema only (no main fallback) for the duration of the SQL body, then restored to <branch>,<main> for subsequent operations. Implemented via pg_catalog.set_config('search_path', ..., true) — the function-call form of SET LOCAL, transaction-scoped and reverted automatically on commit/rollback.

## Tests

New regression test at netbox_branching/tests/test_upgrade.py and a captured 4.4.10 branch fixture at netbox_branching/tests/fixtures/branch_v4_4_10.sql.gz (43KB). The test loads the fixture into a fresh schema and runs branch.migrate() against current NetBox;


arthanson:
@claude /review

RangerRick:
## Pre-PR review — PR #540 (`539-migration` → `main`)

Read-only review. Excludes findings already covered by [Claude's earlier review on this PR](https://github.com/netboxlabs/netbox-branching/pull/540) (concurrency caveat / `database_backwards` not patched / `connection` shadowing / `applied_migrations` comment mismatch / `live_apps` import in hot path / `serialized_rollback` overhead / multi-statement `cursor.execute()` / fixture string-replace safety / `Branch.objects.filter(...).delete()` in `tearDown`).

Verified against `~/git/reference/netbox` at `v4.5.3` (NBE-deployed) and `v4.6.x` to confirm the fix path. The `supports_branching()` resolution covers all 45 migrations across NetBox apps that pair `RunPython` with `apps.get_model` — not just `dcim.0222`. Good blast radius.

---

### P1 — `_fake_for_branch` module import is brittle for non-conventional app layouts

**File:** `netbox_branching/models/branches.py:118`

```python
module = importlib.import_module(f'{migration.app_label}.migrations.{migration.name}')
```

This assumes `migration.app_label` is also the Python package path. That holds for NetBox's first-party apps (`dcim`, `extras`, …) but breaks silently for:

- Plugins where `AppConfig.app_label` differs from `AppConfig.name` (e.g. `name = 'mycompany.netbox_plugin'`, `app_label = 'mycompany_plugin'`).
- Apps using `MIGRATION_MODULES = {'app_label': 'some.other.path'}`.

In both cases `ModuleNotFoundError` is caught and the heuristic silently runs — the user's `fake_on_branch` override is ignored without warning, exactly when they need the escape hatch most.

Django already loaded the migration; reuse what it loaded:

```python
import sys
module = sys.modules.get(migration.__class__.__module__)
if module is not None and (explicit := getattr(module, 'fake_on_branch', None)) is not None:
    return bool(explicit)
```

Same pattern applies at `netbox_branching/models/branches.py:455`.

---

### P2 — `fake_on_branch = False` docs lack a "don't shoot yourself" warning

**File:** `docs/plugin-development.md:175-181`

The "When to set `fake_on_branch = False`" section explains *why* you'd override the heuristic, but doesn't note the failure mode:

> If a migration's only operations target non-branchable models (e.g. `auth.User`, `contenttypes`) and an author sets `fake_on_branch = False`, the migration will run against the branch schema, which doesn't contain those tables, and fail with `relation "..." does not exist`.

Worth a one-line warning callout in that section so plugin authors don't get a confusing PostgreSQL error and waste time tracing it back to the override.

---

### P2 — `test_upgrade_from_v4_4_10` only asserts status, not data correctness

**File:** `netbox_branching/tests/test_upgrade.py:99-107`

```python
branch.migrate(user=user)
branch.refresh_from_db()
self.assertEqual(branch.status, BranchStatusChoices.READY, ...)
```

The whole point of the bug being fixed is that `populate_port_template_mappings` reads `FrontPortTemplate.rear_port_position` and silently writes to the wrong schema. A passing migration that produced empty `PortTemplateMapping` rows on the branch would still satisfy the current assertions. Add at least one data-shape check after `branch.migrate()`:

```python
with connection.cursor() as cursor:
    cursor.execute(f'SET search_path TO "{branch.schema_name}", public')
    cursor.execute('SELECT COUNT(*) FROM dcim_porttemplatemapping')
    self.assertGreater(cursor.fetchone()[0], 0)
```

(Pick whichever count is non-zero in the fixture.) This is the regression assertion that locks in the routing behavior, not just the absence of an exception.

---

### P2 — Failure path is untested

**File:** `netbox_branching/models/branches.py:724-731`

The new `FAILED` status logic is the riskiest behavioral change in the PR — replacing `status = READY; save()` with a targeted `update()` and re-raise — and nothing exercises it. A second test that triggers a deliberate migration failure (e.g. monkey-patch `apply_migration` to raise on the second iteration) and asserts:

1. The exception propagates to the caller.
2. `Branch.objects.get(pk=...).status == FAILED` post-rollback.
3. (Bonus, since it ties to the comment-vs-code mismatch Claude flagged) `applied_migrations` reflects whatever was committed before the failure.

---

### Verified-OK items (not findings, just notes)

- `_branch_isolated_runsql` against NetBox 4.5/4.6 RunSQL ops (`extras.0117_move_objectchange`, `extras.0111_rename_content_types`, `tenancy.0020_remove_contactgroupmembership`, etc.) — all rename branchable tables/indexes that exist in the branch schema, so the restricted `search_path` works correctly.
- `users/migrations/*` RunSQL ops — `users.User` is non-branchable, `_fake_for_branch` correctly fakes the whole migration before isolation matters.
- `main_schema` defaults to `'public'` (`netbox_branching/__init__.py:34`), so the `f'{branch_schema},{main_schema}'` restoration value is well-formed under default config.

arthanson:
Claude reviews - first one no issues, second either outside the scope of this PR (_fake_for_branch possibly worth a separate PR). Test only asserts status not data correctness, would need a more comprehensive test and could break in future if changing schema, right now just want to make sure the migrations pass.

RangerRick:
> Claude reviews - first one no issues, second either outside the scope of this PR (_fake_for_branch possibly worth a separate PR). Test only asserts status not data correctness, would need a more comprehensive test and could break in future if changing schema, right now just want to make sure the migrations pass.

@arthanson yeah, no P0s, just thought it might be useful since it found a few little things not caught by the other pass

jnovinger (approved):
I can't find anything egregious, especially outside of what Claudebot and @RangerRick raised. Nice work.

PR #543, opened 2026-05-07
Fixes #542: Disconnect signal handlers during branch migration

### LINKED ISSUE #542: Data migrations applied to a branch are recorded as user changes
### Plugin Version

1.0.2

### NetBox Version

4.5.3

### Python Version

3.12

### Steps to Reproduce

1. Stand up a clean NetBox 4.5.3 install with netbox-branching 1.0.2 enabled, and run python manage.py migrate.
2. Create a branch (UI or shell):
```python
from netbox_branching.models import Branch
Branch.objects.create(name='repro')
# wait for the provisioning job to finish, branch reaches READY
```
3. From python manage.py nbshell, replicate exactly what Branch.migrate()'s apply loop does — with activate_branch(self): wrapping a single Model.save() on a branchable model:
```
from netbox_branching.models import Branch
from netbox_branching.contextvars import active_branch
from dcim.models import Manufacturer

branch = Branch.objects.get(name='repro')

token = active_branch.set(branch)
try:
    Manufacturer(name='m1', slug='m1').save(using=branch.connection_name)
finally:
    active_branch.reset(token)

print(branch.get_unmerged_changes().count())
```

### Expected Behavior

`branch.get_unmerged_changes().count()` returns `0`. Migration-time writes (data migrations, or anything else done with the branch active outside a user-driven sync) should not appear as user-visible pending changes on the branch.

### Observed Behavior

`branch.get_unmerged_changes().count()` returns `1`. The save fired `post_save` → `core.signals.handle_changed_object`, which — because `active_branch` is set — recorded an `ObjectChange` row in the branch schema. That row then shows up as a pending change.

In the real-world path that motivated this report, applying a NetBox data migration through `MigrateBranchJob` produced ~100 spurious pending changes on the branch (one per object touched by the migration's `RunPython`).

Root cause: `SyncBranchJob` deliberately disconnects the changelog handlers around `branch.sync()` (`netbox_branching/jobs.py:70-77`):
```python
def _disconnect_signal_receivers(self):
    post_save.disconnect(handle_changed_object)
    m2m_changed.disconnect(handle_changed_object)
    pre_delete.disconnect(handle_deleted_object)
    post_clean.disconnect(validate_branching_operations)
```

`MigrateBranchJob.run()` has no equivalent — it just calls `branch.migrate(...)`. Since v1.0.2's `Branch.migrate()` wraps the work in `with activate_branch(self):` (so the router sends writes to the branch schema), every ORM write performed by a data migration also fires the changelog signals against the branch and lands as an `ObjectChange`.

Suggested fix: lift `_disconnect_signal_receivers` / `_reconnect_signal_receivers` into a shared helper (or a context manager) and call it from `MigrateBranchJob.run()` around `branch.migrate(...)`, mirroring `SyncBranchJob`. Same reasoning applies to any other branch-active path that isn't user-driven editing.

jnovinger:
### Fixes: #542

Adds `disconnect_signal_receivers()` context manager to `MigrateBranchJob.run()`, preventing data migrations from creating spurious ObjectChange records in the branch schema. This mirrors the pattern already used by `SyncBranchJob`.

The disconnect/reconnect logic is extracted into a shared context manager for reuse across branch operations.

### DIFF
--- a/netbox_branching/jobs.py
+++ b/netbox_branching/jobs.py
@@ -39,7 +39,7 @@ def get_job_log(job):
 
 
 @contextmanager
-def disconnect_signal_receivers():
+def disconnect_object_change_signal_handlers():
     """
     Context manager to temporarily disconnect object change signal handlers. Used during branch
     operations (sync, migrate) that modify the branch schema but should not record ObjectChange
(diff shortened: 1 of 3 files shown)


arthanson:
@claude /review

RangerRick:
Just passed my runbook reproducing Natalia's issue, this is good to go for me. Once this is out I'll bump NBE to use it and my NBE plugin changes and we'll do another regression pass. Thanks!

RangerRick:
Actually, I take that back. The migrations passed, but now trying to merge back to main is throwing branching errors. Will collect info.

RangerRick:
Looks like it was just a change necessary on our reconciliation side. All good, ready to ship. Thanks!

arthanson (approved):

PR #544, opened 2026-05-08
allow plugins to register include_models

### LINKED ISSUE #545: Allow plugins to register a resolver for branching support on non-ChangeLoggingMixin models
### Plugin Version

v1.0.2

### Proposed functionality

Currently branching has a hard-coded INCLUDE_MODELS that contains things like through models that need to be branch aware but are not include the ObjectChange Mixin.  Should provide an extension mechanism so that plugins can register models to also be included.

### Use case

`INCLUDE_MODELS` is a static constant in netbox-branching and only covers NetBox's own non-ChangeLoggingMixin through tables (`extras.taggeditem`, `dcim.portmapping`, etc.). A plugin that defines its own through tables - whether declared statically or generated at runtime - has no way to add itself to that list, so its through-table queries fall through to the default ChangeLoggingMixin heuristic and get routed to main even inside an active branch. This breaks FK constraints when the parent row exists only in the branch and leaks branch-scoped relationships into main.

A register_branching_resolver() hook lets a plugin opt its own models into branching from PluginConfig.ready(), giving plugins the same capability INCLUDE_MODELS provides for built-in models.

### External dependencies

N/A

arthanson:
### Fixes: netboxlabs/netbox-branching#545

Adds a public extension point register_branching_resolver() so plugins can opt their own models into branching support, parallel to the existing static INCLUDE_MODELS constant used for NetBox's built-in non-ChangeLoggingMixin models.

### DIFF
--- a/docs/plugin-development.md
+++ b/docs/plugin-development.md
@@ -71,6 +71,90 @@ exempt_models = ['my_plugin.*']
 
 See [Configuration: `exempt_models`](configuration.md#exempt_models) for full details.
 
+## Opting In: `register_branching_resolver`
+
+Some plugins have models that are not `ChangeLoggingMixin` subclasses but still need to participate in branching — most commonly **dynamically-generated M2M through tables** that store relationships involving branchable parent objects.
+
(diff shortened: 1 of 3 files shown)


arthanson:
@claude /review

jnovinger:
Closing in favor of #545 .

PR #548, opened 2026-05-11
Fixes #545: Allow plugins to register branching resolver

### LINKED ISSUE #545: Allow plugins to register a resolver for branching support on non-ChangeLoggingMixin models

jnovinger:
### Fixes: #545

Adds a public extension point `register_branching_resolver()` so plugins can opt their own models into branching support, parallel to the existing static `INCLUDE_MODELS` constant used for NetBox's built-in non-ChangeLoggingMixin models.

This is a retarget of #544 (which targeted `main`) onto `feature` for inclusion in the next minor release. Commits are cherry-picked from Arthur's original branch; `main` was merged into `feature` first to ensure a clean apply.

### DIFF
--- a/docs/plugin-development.md
+++ b/docs/plugin-development.md
@@ -71,6 +71,90 @@ exempt_models = ['my_plugin.*']
 
 See [Configuration: `exempt_models`](configuration.md#exempt_models) for full details.
 
+## Opting In: `register_branching_resolver`
+
+Some plugins have models that are not `ChangeLoggingMixin` subclasses but still need to participate in branching — most commonly **dynamically-generated M2M through tables** that store relationships involving branchable parent objects.
+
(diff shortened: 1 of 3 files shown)


jnovinger:

Note that this is a simple re-targeting of this PR from `main` to `feature`. You previously reviewed this work and gave it a 👍 in https://github.com/netboxlabs/netbox-branching/pull/544#issuecomment-4410601802. This review should be a formality.

jeremystretch (approved):

PR #550, opened 2026-05-12
#567 - Make ObjectChange field-key handling extensible for replay and conflict detection

arthanson:
### Fixes: #567 

Introduces an internal extension point that makes the field-name keys in an ObjectChange data dict translatable before branching applies or compares them, so renames in a model's runtime schema don't cause silently-dropped writes or spurious conflicts during sync, merge, and revert.

## Changes

- `netbox_branching/utilities.py` — add `register_objectchange_field_migrator()` and `resolve_objectchange_field_migration()`. `update_object()` runs data through the resolver once  before the apply loop.
- `netbox_branching/models/changes.py` - `ChangeDiff._update_conflicts()` runs original, modified, and current through the resolver before comparison.
- `netbox_branching/tests/test_objectchange_field_migrator.py` - covers registration validation, both call sites, first-non-None-wins ordering, key-dropping, and error isolation.

**Note:** The hook is currently internal and subject to change; default behavior is unchanged when no migrator is registered.


arthanson:
@claude /review

arthanson:
Claude review:

1. Acknowledged in a comment above the cached properties - display reads raw on purpose; migrating here without also updating views.py's raw-dict lookups would KeyError on translated keys.
2. Pre-existing pattern, leaving as-is: a misbehaving migrator should fail loudly (KeyError) rather than silently emit spurious conflicts via .get() returning None.
3. Removed from __all__ - kept register_objectchange_field_migrator since external callers need it, dropped the resolver since it's internal-only.

jeremystretch:

netbox_branching/models/changes.py
```diff
@@ -199,29 +203,44 @@ def get_action_color(self):
...
-                k for k, v in self.original.items()
-                if v != self.current.get(k)
+                k for k, v in original.items()
+                if v != current.get(k)
             ]
         self.conflicts = conflicts or None
 
+    # Display properties below read raw dicts; migration is only applied where
+    # snapshots are compared (``_update_conflicts``).  If migrating here, also
+    # update the raw-dict lookups in ``views.py`` to avoid KeyErrors.
```
jeremystretch: I found this comment a bit confusing, like maybe it was left over from an implementation decision. Unsure what "If migrating here" signals.

jeremystretch (changes requested):
One minor nit. Also check Claude's feedback above but seems like it can be safely ignored.

PR #551, opened 2026-05-12
change last_updated to auto_now=True

### LINKED ISSUE #483: The `last_updated` field on ChangeDiff should specify `auto_now=True`
### Plugin Version

0.8.3

### NetBox Version

4.5.5

### Python Version

3.12

### Steps to Reproduce

The `last_updated` time of ChangeDiff objects is not being updated appropriately, because the field is declared with `auto_now_add=True` rather than `auto_now=True`.

### Expected Behavior

`last_updated` should be updated automatically whenever a ChangeDiff instance is updated.

### Observed Behavior

`last_updated` is set only when the ChangeDiff instance is created.

arthanson:
### Fixes: #483 

 ChangeDiff.last_updated was declared auto_now_add=True, so the timestamp was set only at row creation and never refreshed on subsequent saves - despite the field's name and its use as the default ordering key (ordering = ('-last_updated',)).

Flips the field to auto_now=True, adds the corresponding AlterField migration, and drops two now-redundant timezone.now() assignments at the Model.save() call sites in signal_receivers.py. The third site (line 136) uses QuerySet.update(), which bypasses auto_now, so its explicit assignment is left in place.

Existing rows keep their old last_updated until the next save (AlterField does not backfill); this is intentional.

Adds a regression test that back-dates last_updated via .update() and asserts the next save() advances it.

### DIFF
--- a/netbox_branching/migrations/0009_changediff_last_updated_auto_now.py
+++ b/netbox_branching/migrations/0009_changediff_last_updated_auto_now.py
@@ -1,0 +1,16 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('netbox_branching', '0008_branch_custom_permissions'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='changediff',
+            name='last_updated',
+            field=models.DateTimeField(auto_now=True),
+        ),
+    ]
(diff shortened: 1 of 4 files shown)


arthanson:
@claude /review

jeremystretch (approved):

PR #553, opened 2026-05-13
#552 add merge/revert to 4.10 migration test

### LINKED ISSUE #552: Add branching test to merge/revert migrated data
### Proposed Changes

There is a test_upgrade script that currently migrates data from a 4.10 netbox install and makes sure it migrates.  Tests should be added to merge and then revert this data.

### Justification

The migration of this data checks that it migrated, but doesn't check that it migrate correctly and is actually usable.  Doing a merge and revert would more completely test that the migration worked correctly and that all the ObjectChange records are correct - which was an issue previously so should be tested.

arthanson:
### Fixes: #552

Adds merge and revert test to test_upgrade, needed a couple change - the fixture needed to be updated to include the main database (for content_types) not just the branch schema that it had before, this also required the test to actually reload the test database as the normal test database doesn't have the same content-types and such that the fixture requires - so it creates a new test database from the fixture.


arthanson:
@claude /review

arthanson:
Claude review issues 1-4 addressed, 4 and 5 are non issues.

jeremystretch (commented):
This appears to have broken several other tests. I suspect the test database isn't being cleaned up properly.

PR #556, opened 2026-05-13
#383 document obj.snapshot() if modifying data in scripts

### LINKED ISSUE #383: When updating data via Script, Changes Ahead tab displays all fields from before as None
### Plugin Version

0.7.1

### NetBox Version

4.4.8

### Python Version

3.10.2

### Steps to Reproduce

1. Create custom script.
2. Run it
3. Commit changes
4. Got to branch and Changes Ahead tab

### Expected Behavior

Let's say script made update to interface description from nothing to "Test", then in changes tab I should see only description update only to said interface.

Image

### Observed Behavior

But instead it sais that every field was None and then change to something.

Image

arthanson:
### Fixes: #383

This turned out to be a documentation issue - in Scripts if you would need to call obj.snapshot() to get the pre-change data correct. This is what NetBox does internally.

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -79,7 +79,7 @@ Here are recommended best practices for working with NetBox branches:
 
 When an object is modified through NetBox's web UI or REST API, NetBox automatically captures a "before" snapshot of the object so the resulting changelog entry has both pre- and post-change data. **Code that bypasses the views and viewsets — Custom Scripts, `nbshell` sessions, plugin background jobs, bulk-import helpers — does not get this snapshot automatically.**
 
-Without a prechange snapshot, the resulting `ObjectChange` has empty `prechange_data`. This is mostly invisible in main NetBox (the changelog UI infers a "before" state from the prior change), but in a branch it has two consequences:
+Without a pre-change snapshot, the resulting `ObjectChange` has empty `prechange_data`. This is mostly invisible in main NetBox (the changelog UI infers a "before" state from the prior change), but in a branch it has two consequences:
 
 1. The **Changes Ahead** diff cannot render a real before/after comparison for the affected object.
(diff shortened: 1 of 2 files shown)


docs/best-practices.md
```diff
@@ -75,6 +75,26 @@ Here are recommended best practices for working with NetBox branches:
 
 * **Conflict Avoidance:** Be mindful of potential conflicts when multiple branches update the same data.
 
+### Programmatic Modifications: Scripts, Shells, and Custom Code
+
+When you modify an object through NetBox's web UI or REST API, NetBox automatically captures a "before" snapshot of the object so the resulting changelog entry has both prechange and postchange data. **Code that bypasses the views and viewsets — Custom Scripts, `nbshell` sessions, plugin background jobs, bulk-import helpers — does not get this snapshot for free.**
```
jeremystretch:
```suggestion
When an object is modified through NetBox's web UI or REST API, NetBox automatically captures a "before" snapshot of the object so the resulting changelog entry has both pre- and post-change data. **Code that bypasses the views and viewsets — Custom Scripts, `nbshell` sessions, plugin background jobs, bulk-import helpers — does not get this snapshot automatically.**
```

docs/best-practices.md
```diff
@@ -75,6 +75,26 @@ Here are recommended best practices for working with NetBox branches:
...
+
+Without a prechange snapshot, the resulting `ObjectChange` has empty `prechange_data`. This is mostly invisible in main NetBox (the changelog UI infers a "before" state from the prior change), but in a branch it has two consequences:
```
jeremystretch:
```suggestion
Without a pre-change snapshot, the resulting `ObjectChange` has empty `prechange_data`. This is mostly invisible in main NetBox (the changelog UI infers a "before" state from the prior change), but in a branch it has two consequences:
```

jeremystretch (approved):
Approved with some suggested tweaks

PR #557, opened 2026-05-13
#530 call super().clean() in Branch.clean()

### LINKED ISSUE #530: Branch.clean() does not call super().clean(), preventing CUSTOM_VALIDATORS
### Plugin Version

0.9.1

### NetBox Version

4.5.9

### Python Version

3.13

### Steps to Reproduce

1. Define a custom validator for the Branch model in `configuration.py`:

```python
from extras.validators import CustomValidator

class BranchNameValidator(CustomValidator):
    def validate(self, instance, request=None):
        if not instance.name.startswith('TICKET-'):
            self.fail('Branch name must start with TICKET-', field='name')

CUSTOM_VALIDATORS = {
    'netbox_branching.branch': [BranchNameValidator()],
}
```

2. Create a new branch with a name that does not match the validator (e.g. `my-test-branch`)

### Expected Behavior

Branch creation should fail with a ValidationError indicating the name must start with `TICKET-`.

### Observed Behavior

Branch is created without any validation error. The custom validator is never invoked.

`Branch.clean()` does not call `super().clean()`, which means `CustomValidationMixin.clean()` never fires. That method is responsible for sending the `post_clean` signal, which triggers `CUSTOM_VALIDATORS` lookup and execution.

This was introduced in ea5ea272 (September 2024) when `clean()` was re-added to enforce `max_branches` without restoring the `super().clean()` call that had existed in the earlier `Context` model.

arthanson:
### Fixes: #530

Branch.clean() should call super().clean()

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -264,6 +264,7 @@ class Branch(JobsMixin, PrimaryModel):
         return f'schema_{self.schema_name}'
 
     def clean(self):
+        super().clean()
 
         # Enforce the maximum number of total branches
         if not self.pk and (max_branches := get_plugin_config('netbox_branching', 'max_branches')):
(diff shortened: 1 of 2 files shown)


arthanson:
@claude /review

jeremystretch (approved):

PR #571, opened 2026-05-22
Expose signal to extend SquashMergeStrategy dependency graph

arthanson:
### Fixes: #572 

Adds a squash_dependency_graph_built Django signal to netbox_branching.merge_strategies.squash, fired by SquashMergeStrategy._order_collapsed_changes() immediately after _build_fk_dependency_graph() populates the FK/GFK edges and before _dependency_order_by_references() produces the topological order.

### Signal contract:
- sender: SquashMergeStrategy
- collapsed_changes (kwarg): the dict of CollapsedChange instances keyed by (app_label.model_name, pk) that is about to be ordered. Receivers mutate each entry's depends_on / depended_by sets in place to add edges; the return value is ignored.

The signal is exported via __all__ alongside SquashMergeStrategy. No behavior change when no receivers are connected - default merge ordering is preserved.

Also contains a fix for the recent NetBox change to TaggableManager the was effecting tests.


jnovinger:

Ignore the failing tests in CI for the moment, those are from an unrelated change introduced in upstream NetBox.

jnovinger (commented):
> #### 1. Signal fires during `revert()` as well as `merge()`
> 
> `squash_dependency_graph_built` is fired inside `_order_collapsed_changes()`, which is called by both `merge()` and `revert()` (line 258 of `squash.py`). Receivers have no way to distinguish which operation is in progress. A receiver that adds dependency edges to enforce merge-time ordering may produce incorrect results during revert — even though revert simply reverses the topological order, the added edges will still participate in that sort and could pull changes into the wrong position.

I think this one is worth thinking about. Was this the intended design?

arthanson:
changed 1 and moved for 2.  1 isn't really used by custom objects but added it for future flexibility.

netbox_branching/merge_strategies/squash.py
```diff
@@ -684,6 +694,11 @@ def _order_collapsed_changes(collapsed_changes, logger):
         )
 
         SquashMergeStrategy._build_fk_dependency_graph(deletes, updates, creates, logger)
+        squash_dependency_graph_built.send(
+            sender=SquashMergeStrategy,
+            collapsed_changes=to_process,
+            operation=operation,
+        )
```
jnovinger, line 701: It appears that `_split_bidirectional_cycles()` can create synthetic UPDATE entries with 3-tuple keys like ('app.model', pk, 'update_fieldname') instead of the usual 2-tuple ('app.model', pk). The `collapsed_changes` dict passed to receivers will contain both formats. Worth either documenting in the signal's docstring or normalizing the key format so receivers don't need to handle both. What do you think?

jnovinger (commented):
Looks good overall. A few things worth keeping in mind as custom-objects builds the receiver for this (none blocking):

- CollapsedChange isn't exported from `merge_strategies/__init__.py`, so the receiver will need to import from `merge_strategies.squash` directly. Fine for now since we control both sides, but worth remembering if this ever becomes a public extension point.
- If a receiver adds a `depends_on` edge referencing a key that doesn't exist in `collapsed_changes`, the topological sort could report a false cycle with no indication the receiver caused it. Something to watch for during receiver development.
- The signal only fires for squash strategy, not iterative, correct? Might be worth documenting somewhere if we want receivers to be aware of that constraint.

netbox_branching/merge_strategies/squash.py
arthanson: documented this, outside the scope of this PR to make a larger change like this, although potentially worth doing it in the future.

arthanson:
> * CollapsedChange isn't exported from `merge_strategies/__init__.py`, so the receiver will need to import from `merge_strategies.squash` directly. Fine for now since we control both sides, but worth remembering if this ever becomes a public extension point.

noted - for internal use only right now.

> * If a receiver adds a `depends_on` edge referencing a key that doesn't exist in `collapsed_changes`, the topological sort could report a false cycle with no indication the receiver caused it. Something to watch for during receiver development.

yes - if we make it more publicly accessible will need to make it more robust and documented.

> * The signal only fires for squash strategy, not iterative, correct? Might be worth documenting somewhere if we want receivers to be aware of that constraint.

only one currently needed - this isn't a complete generalized signal handler to be used by plugin maintainers, it is a surgical change for what is needed currently.  If we make it public in the future would need a more generalized implementation on both merge strategies.

jnovinger (approved):

jnovinger (changes requested):
CI failure

PR #577, opened 2026-05-28
#28 - Preserve sync conflict resolution when merging branch

### LINKED ISSUE #28: Conflicting changes from a sync should be retained when merging a branch
### Plugin Version

v0.2.0

### NetBox Version

v4.1-beta1

### Python Version

3.10

### Steps to Reproduce

1. Create a branch
2. Within main, change the status of a site from active to staging
3. Activate the branch
4. Change the status of that same site from active to planned
5. Synchronize the branch
6. Acknowledge the conflict and commit changes

### Expected Behavior

The value in the branch should be updated to reflect that of main, and its changelog should be updated accordingly.

### Observed Behavior

Within the branch, the site's status now shows "staging." However, the branch's change history (under the "changes ahead" tab) shows that the site's status will be changed to "planned" when merging the branch.

arthanson:
### Fixes: #28 

For a conflict, instead of changing the existing ObjectChange as suggested in the issue, this creates a new ObjectChange with the updated change. This keeps the ObjectChange immutable and matches what we do elsewhere.  It makes it so there is no conflict shown anymore and has a changelog message that it comes from the sync.

During sync, for each non-DELETE change from main against an object the branch has also touched:

- Capture the branch's pre-sync state of the object
- Apply main's change to the branch schema
- Capture the branch's post-sync state
- If the state changed, write a synthetic ObjectChange (prechange = pre-sync, postchange = post-sync) into the branch schema
- Create it via the core ObjectChange model (not the proxy) so record_change_diff fires and clears the acknowledged conflict on the existing ChangeDiff

Merge replay (iterative and squash) then sees the synthetic update last and lands on main's synced value.

### DIFF
--- a/netbox_branching/models/branches.py
+++ b/netbox_branching/models/branches.py
@@ -566,18 +566,22 @@ class Branch(JobsMixin, PrimaryModel):
     # Branch actions
     #
 
-    def _apply_sync_update(self, change, user, logger, request_id, touched_object_keys):
+    def _apply_sync_update(self, change, logger, touched_object_keys, sync_buffer):
         """
         Apply a non-DELETE change from main and, when the branch has also touched the
-        same object, record a sync-originated ObjectChange so the merge replay does
-        not undo it (#28).
+        same object, buffer the pre/post state so a single synthetic ObjectChange can
+        be written per object after the sync loop completes (#28).
 
         ``touched_object_keys`` is a set of ``(content_type_id, object_id)`` tuples
         for which the branch has at least one ChangeDiff. It is pre-fetched once by
         the caller to avoid an N+1 query inside the sync loop.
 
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
Claude comments:

- **1:** atomic span: skipped — current design is necessary for dry-run correctness, alternatives don't preserve commit=False semantics.
- **2:** local import: moved ChangeDiff to the top-level from .changes import ChangeDiff, ObjectChange and removed the inline import.
- **3:** multi-synthetic docstring: added a paragraph to _apply_sync_update's docstring explaining that sequential main changes produce multiple synthetic entries, each capturing only its delta, with merge replay landing on main's final value.
- **4:** format contract: no code change — shared risk with existing patterns, called out for awareness only.

jnovinger (approved):

PR #591, opened 2026-06-05
Performance Enhancements to branch provisioning

### LINKED ISSUE #595: Parallelize branch provisioning to reduce time on large databases
### Plugin Version

v1.1.0

### Proposed functionality

Speed up branch provisioning by performing the table copy and index builds in parallel rather than serially, while producing a branch schema identical to what provisioning creates today.

### Use case

On installs with multi-GB branchable data, provisioning is the dominant cost of creating a branch and can take many minutes, during which the branch is unusable. Reducing this time makes branching more practical at scale.

### External dependencies

N/A

arthanson:
### Fixes: #595 

Building indexes after a bulk INSERT is much faster than maintaining them row-by-row during the copy - each index gets built once from sorted data, and the builds parallelize. 

So Phase 1 creates empty tables, Phase 2 streams the data in under an exported MVCC snapshot, and Phase 3 replays the indexes and constraints against the populated tables in parallel.

When streaming the data it does largest tables first as they take the most time the smaller tables can be done by the other workers quickly. Parallelism is set by provision_workers (default 4).

A fourth pass runs ANALYZE so the planner has stats immediately.

With very large database - timing of provisions a branch went from 18 minutes down to 4 minutes with these changes.


arthanson:

arthanson:

arthanson:
per claude review:

1. not a valid case/ comment, if you are running migrations you are upgrading the system.  Removed the doc comment on this.
2.  3. 4. fixed
5. added the new test case.

netbox_branching/jobs.py
```diff
@@ -58,7 +58,28 @@ def disconnect_object_change_signal_handlers():
...
+    Base class for branch operation jobs.
+
+    Branch operations (provisioning, sync, merge, revert, migrate) routinely run longer than RQ's
+    default job timeout (NetBox's RQ_DEFAULT_TIMEOUT, 300s), at which point RQ kills the worker with
+    a JobTimeoutException. NetBox's JobRunner.enqueue() does not consult any per-class timeout, so the
+    configured `job_timeout` must be passed through to RQ at enqueue time. Injecting it here ensures
+    every branch job honours the setting regardless of which call site enqueues it.
+    """
+    @classmethod
+    def enqueue(cls, *args, **kwargs):
```
jeremystretch, line 72: I was going to suggest just defining a `job_timeout` property on this class (like under `SyncBranchJob`) instead of overriding `enqueue()`, but it seems like that doesn't get picked up by the parent `JobRunner`. Should we open a separate bug against NetBox to capture this? (I'm fine with keeping this approach as-is for now.)

netbox_branching/__init__.py
```diff
@@ -68,6 +73,17 @@ def ready(self):
...
+        # Validate provision_workers up front rather than letting a bad value surface as an
+        # unhandled error only when a branch is first provisioned.
+        workers = get_plugin_config('netbox_branching', 'provision_workers')
+        if workers is not None:
+            try:
+                int(workers)
+            except (TypeError, ValueError):
+                raise ImproperlyConfigured(
+                    "netbox_branching: 'provision_workers' must be an integer."
+                )
```
jeremystretch: Suggest getting a little tighter on validation.
```suggestion
            if type(workers) is not int:
                raise ImproperlyConfigured(
                    "netbox_branching: 'provision_workers' must be an integer."
                )
            if workers < 1:
                raise ImproperlyConfigured(
                    "netbox_branching: 'provision_workers' must be greater than or equal to 1."
                )
```

netbox_branching/provisioning.py
```diff
@@ -0,0 +1,470 @@
...
+        for future in as_completed(futures):
+            # worker() funnels task failures into `errors`; this only surfaces an
+            # unexpected error escaping worker() itself.
+            future.result()
+
+    if errors:
+        raise errors[0]
+
+
+def build_main_index_map(cursor, main_schema):
```
jeremystretch, line 201: It would be helpful to document (in the docstring) the arguments for this and the other public functions below.

netbox_branching/provisioning.py
```diff
@@ -0,0 +1,470 @@
...
+
+    def make_copy_task(table):
+        def copy(cursor):
+            cursor.execute("BEGIN")
+            cursor.execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ")
+            cursor.execute(f"SET TRANSACTION SNAPSHOT '{snapshot_token}'")
+
+            main_table = f'{quote_ident(main_schema)}.{quote_ident(table)}'
+            schema_table = f'{quote_ident(schema)}.{quote_ident(table)}'
+            logger.debug(f'Copying {main_table} -> {schema_table}')
```
jeremystretch: Very minor nit, but could we tweak this to log the un-escaped table names? Right now it logs like this:

```
"Copying \"public\".\"dcim_frontport\" -> \"branch_w5aly85j\".\"dcim_frontport\""
```

(All the other loggers are fine.)

jeremystretch (commented):
We should probably re-target this for `feature` given the substantial changes to the provisioning logic and effective job timeouts.

Amazing work @arthanson!

netbox_branching/jobs.py
arthanson: yeah - its sort of a missing feature of the current implementation, we could add something like:

```
class JobRunner(ABC):
      job_timeout = None  # or a classproperty, mirroring `name`

      @classmethod
      def enqueue(cls, *args, **kwargs):
          name = kwargs.pop('name', None) or cls.name
          kwargs.setdefault('job_timeout', cls.job_timeout)
          return Job.enqueue(cls.handle, name=name, *args, **kwargs)
```
will open a FR, but can go ahead with this here.

PR #592, opened 2026-06-08
#588 - Preserve unmodified JSON dict keys (e.g. custom_fields) when merging branches

### LINKED ISSUE #588: Merge incorrect: JSON property override e.g. customfields or local context
### Plugin Version

1.0.3

### NetBox Version

4.6.2

### Python Version

3.12

### Steps to Reproduce

1. Create an object
2. Create and assign two custom fields to the object type
3. Edit the object and set cf1 = 1
4. Create a branch
5. Activate branch
6. Edit the object again, this time set cf2 = 2
7. Inspect the branch diff and see that it shows both customfields set to their values
Image
8. Merge the branch
9. Inspect changelog of the device after merge
Image

### Expected Behavior

Expected behavior would be that in the main, the state would be the same as in the branch.
Setting/having both cf to their value.

### Observed Behavior

Only the changed cf has its value set, instead of both values.

This behavior started with the upgrade of netbox 4.6.x.
In https://github.com/netbox-community/netbox/pull/21691 (merged with 4.6.0) there was a change to better visualize customfield changes by using a deepdiff instead of a shallow diff.
In this change the ObjectChange diff method was changed to provide this deepdiff.
And in the netbox-branching ObjectChange class, to update the object the diff method is used but the changes will be overriden like before 4.6.x. That results in an override of the customfields. 
https://github.com/netboxlabs/netbox-branching/blob/main/netbox_branching/models/changes.py#L83

arthanson:
### Fixes: #588 

NetBox 4.6 (PR #21691) changed ObjectChange.diff() to use deep_compare_dict, so JSON fields like custom_field_data now arrive as partial dicts containing only the keys that changed. update_object() was still setattr-ing those partial dicts, wiping out keys the branch never touched (e.g. setting cf2 in a branch erased cf1 on merge).

Fix is to deep-merge incoming dicts into the existing attribute instead of replacing them.

### DIFF
--- a/netbox_branching/models/changes.py
+++ b/netbox_branching/models/changes.py
@@ -43,6 +43,23 @@ class ObjectChange(ObjectChange_):
         for migrator in branch.migrators.get(object_type, []):
             migrator(self, revert)
 
+    def get_merge_data(self, reverse=False):
+        """
+        Build a deletion-aware partial-update payload for replaying this change's UPDATE
+        onto another schema. By default the payload transforms the pre-change state into
+        the post-change state (used by ``apply()``); with ``reverse=True`` it goes the
+        other way (post -> pre), used by ``undo()``.
+
+        This is intentionally separate from ``diff()``: ``diff()`` is display-oriented
+        (rendered in the change-log tables) and represents a removed key as ``None``,
+        whereas merging needs to drop removed nested JSON keys entirely. ``diff_for_merge``
+        tracks those removals with the ``DELETED`` sentinel so ``update_object`` can apply
+        them rather than leave stale ``None`` values behind. (#588, #592)
+        """
+        if reverse:
+            return diff_for_merge(self.postchange_data_clean, self.prechange_data_clean)
+        return diff_for_merge(self.prechange_data_clean, self.postchange_data_clean)
+
     def apply(self, branch, using=DEFAULT_DB_ALIAS, logger=None, skip_missing=False):
         """
         Apply the change using the specified database connection.
@@ -81,10 +98,9 @@ class ObjectChange(ObjectChange_):
             try:
                 instance = model.objects.using(using).get(pk=self.changed_object_id)
                 logger.debug(f'Updating {model._meta.verbose_name} {instance}')
-                # Build a deletion-aware partial-update payload (pre -> post) rather than
-                # using diff()['post'], which cannot represent removed nested JSON keys. (#592)
-                data = diff_for_merge(self.prechange_data_clean, self.postchange_data_clean)
-                update_object(instance, data, using=using)
+                # Use a deletion-aware payload (pre -> post) rather than diff()['post'],
+                # which cannot represent removed nested JSON keys. (#592)
+                update_object(instance, self.get_merge_data(), using=using)
             except model.DoesNotExist:
                 if skip_missing:
                     logger.debug(f'{model._meta.verbose_name} ID {self.changed_object_id} already deleted; skipping')
@@ -126,8 +142,7 @@ class ObjectChange(ObjectChange_):
         elif self.action == ObjectChangeActionChoices.ACTION_UPDATE:
             instance = model.objects.using(using).get(pk=self.changed_object_id)
             # Reverse direction (post -> pre); keys the branch added are removed. (#592)
-            data = diff_for_merge(self.postchange_data_clean, self.prechange_data_clean)
-            update_object(instance, data, using=using)
+            update_object(instance, self.get_merge_data(reverse=True), using=using)
 
         # Restoring a deleted object
         elif self.action == ObjectChangeActionChoices.ACTION_DELETE:
--- a/netbox_branching/models/changes.py
+++ b/netbox_branching/models/changes.py
@@ -14,6 +14,7 @@ from utilities.serialization import deserialize_object
 
 from netbox_branching.utilities import (
     clear_mptt_fields,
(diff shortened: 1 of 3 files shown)


arthanson:

arthanson:
Per claudes review. 1 is a non issue.  Fixed 2.

arthanson:

arthanson:
1 fixed.

Comment 2 - Invalid: get_clean_data returns getattr(...) or {}, so those properties are never None; the or {} guard is dead code.

Comment 3 - Invalid: snapshot() deep-copies via JSON serialization and save() persists all fields, so the in-place merge is safe and the setattr is a no-op.

netbox_branching/models/changes.py
```diff
@@ -80,7 +81,10 @@ def apply(self, branch, using=DEFAULT_DB_ALIAS, logger=None, skip_missing=False)
             try:
                 instance = model.objects.using(using).get(pk=self.changed_object_id)
                 logger.debug(f'Updating {model._meta.verbose_name} {instance}')
-                update_object(instance, self.diff()['post'], using=using)
+                # Build a deletion-aware partial-update payload (pre -> post) rather than
+                # using diff()['post'], which cannot represent removed nested JSON keys. (#592)
+                data = diff_for_merge(self.prechange_data_clean, self.postchange_data_clean)
```
jeremystretch: Can we move this logic into `ObjectChange.diff()`? It would be best to ensure the diff data is generated consistently. (If not, we should remove the `diff()` method as it's no longer used anywhere.)

netbox_branching/utilities.py
```diff
@@ -350,6 +350,94 @@ def clear_mptt_fields(instance):
...
+
+    def __repr__(self):
+        return '<DELETED>'
+
+
+# Singleton sentinel instance, compared by identity.
+DELETED = _DeletedKey()
+
+
+def diff_for_merge(source, destination):
```
jeremystretch: Why is this split into two functions?

netbox_branching/utilities.py
```diff
@@ -350,6 +350,94 @@ def clear_mptt_fields(instance):
...
+    Compute the partial-update payload that transforms ``source`` into ``destination``
+    when deep-merged onto a target via ``update_object`` / ``_deep_merge_dict``.
+
+    This mirrors the "added" side of NetBox's ``deep_compare_dict`` — only differing
+    keys are returned, and nested dicts yield only their changed keys so that keys the
+    branch never touched are preserved on the target schema (#588) — with one critical
+    difference: a key present in ``source`` but absent from ``destination`` *within a
+    nested dict* is represented with the ``DELETED`` sentinel instead of being dropped.
+
+    ``deep_compare_dict`` reads nested values with ``dict.get()``, so it cannot tell
```
jeremystretch, line 381: This sounds like a bug in NetBox. Shouldn't `deep_compare_dict()` be setting keys to None in the change data? If my understanding is correct, it seems like we should fix this in NetBox rather than working around it in the plugin.

netbox_branching/utilities.py
```diff
@@ -350,6 +350,94 @@ def clear_mptt_fields(instance):
...
+    merged key-by-key; the ``DELETED`` sentinel removes the key from ``target``;
+    all other values are replaced.
+    """
+    for key, value in source.items():
+        if value is DELETED:
+            target.pop(key, None)
+        elif isinstance(value, dict) and isinstance(target.get(key), dict):
+            _deep_merge_dict(target[key], value)
+        else:
+            target[key] = value
```
jeremystretch: Per Claude:

If `value` is a nested dict that contains a `DELETED` sentinel but the corresponding slot in target is **not** a dict (None / scalar / missing), the sentinel-bearing dict is assigned verbatim. A `_DeletedKey` instance is not JSON-serializable, so the subsequent `save()` would raise rather than corrupt data — but it's still a crash.

This requires main's current value at a nested key to have diverged from the branch's prechange snapshot (e.g. main has `custom_field_data = {'jsoncf': None}` while the branch removed a key inside `jsoncf`). Conflict detection blocks many such cases at merge time, but the revert path and partially-synced states are harder to reason about — `local_context_data` in particular is frequently None. Worth closing defensively:

```
else:
    target[key] = _strip_deleted(value)
```

`_strip_deleted` is a no-op on non-dicts, so this is safe for all values.

netbox_branching/utilities.py
jeremystretch: Remember to add this to `__all__`.

jeremystretch (changes requested):
If I understand the root problem correctly, this seems like something we should fix in NetBox itself rather than in the plugin, by ensuring keys set to None are reflected accurately by `deep_compare_dict()`.

netbox_branching/utilities.py
arthanson: The lossiness does originate in NetBox, but I don't think we need a NetBox fix. `diff_for_merge` works off the raw `prechange_data_clean`/`postchange_data_clean` snapshots, not `diff()` - and those already carry the ground truth (a removed key is just absent from the post snapshot), so we reconstruct the correct delta locally with no dependency on NetBox.

Changing `deep_compare_dict` wouldn't help much.  A real fix is a sentinel rewrite touching the change-log UI, a public contract, and existing tests - and even then it'd only let us drop the small diff helper, not the merge logic.

jeremystretch (approved):

PR #597, opened 2026-06-11
496 skip the branch deleted-in-main check when there is no request

### LINKED ISSUE #496: Custom scripts cannot change interfaces on devices created by itself in a branch
### Plugin Version

0.8.3

### NetBox Version

v4.5.5 (docker image netboxcommunity/netbox:v4.5.5)

### Python Version

3.12.3

### Steps to Reproduce

0. Create a fresh NetBox install
1. Create a site
2. Create a device role with the name 'switch'
3. Create a device template with at least one interface
4. Add this script to NetBox: https://github.com/GoepiGB/netbox-branching-issue-496/blob/main/BranchingTest.py
5. Create a branch
6. Run the script; select the site and device template created before, committing the changes is not necessary

### Expected Behavior

The device is created, and the interface is assigned a description successfully

### Observed Behavior

The script fails with a validation error:

`ValidationError: {'__all__': ["Cannot modify interface '1' because it has been deleted in the main branch. Sync with the main branch to update."]}`

Image

Running the script in the main branch works without error.

arthanson:
### Fixes: #496 

A custom script run with commit disabled creates objects in a branch and then editing one of those interfaces fails with a false Cannot modify … because it has been deleted in the main branch.

## Cause

In dry-run NetBox skips the event_tracking request processor, so current_request is never set. Without a request, core.signals.handle_changed_object writes no ObjectChange, and therefore the plugin records no ChangeDiff. validate_branching_operations decides "was this created in the branch?" entirely from ChangeDiff, so with none present it mistakes the just-created object for one deleted in main.

## Fix

Validate_branching_operations returns early when current_request is unset - the same condition NetBox uses to gate change logging. With no request there are no ChangeDiffs to evaluate and nothing is being persisted anyway, so the check stands down instead of false-positiving. The deleted-in-main guard still fires for real requests and committing runs (the #422 test is updated to set a request, matching that real scenario).

### DIFF
--- a/netbox_branching/signal_receivers.py
+++ b/netbox_branching/signal_receivers.py
@@ -13,6 +13,7 @@ from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from extras.events import process_event_rules
 from extras.models import EventRule
+from netbox.context import current_request
 from netbox.signals import post_clean
 from utilities.exceptions import AbortRequest
 from utilities.serialization import serialize_object
@@ -79,6 +80,12 @@ def validate_branching_operations(sender, instance, **kwargs):
 
     # Only validate if we're in a branch and this model supports branching
     if branch is None:
+        return
+
+    # Without a request, NetBox writes no ObjectChange (and so no ChangeDiff), so the
+    # ChangeDiff-based accessibility check below has nothing to read and would
+    # false-positive on objects the branch just created. See issue #496.
+    if current_request.get() is None:
         return
 
     # Check if this model supports branching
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
re 1 - not valid, Branch-schema existence can't tell a branch-created object from a main-deleted one (both exist in the isolated schema), so it'd disable the #422 guard; the CREATE ChangeDiff is the only reliable signal and it's absent without a request.

item 2 fixed.

jeremystretch (approved):

PR #599, opened 2026-06-15
#581 - Close branch connections during post-migrate pending-migrations check

### LINKED ISSUE #581: `check_pending_migrations` causes Postgres OOM with many open branches
**Plugin Version:** v1.0.3 (also reproducible on v0.9.1)

**NetBox Version:** v4.5.7

**Python Version:** 3.12

**Steps to Reproduce:**

1. Have 75+ branches in `READY` status
2. Install or upgrade a plugin that introduces new database migrations
3. Run `manage.py migrate`
4. All migrations apply successfully against the main schema
5. After all apps finish migrating, Django fires the `post_migrate` signal for each app. When `sender.name == 'netbox_branching'` (the last app to signal), `check_pending_migrations` runs
6. The handler iterates all ready branches; for each, `branch.pending_migrations` instantiates a `MigrationExecutor`, which runs `introspection.table_names()` against the branch schema

**Environment:**

| Parameter | Value |
|---|---|
| Postgres | 16 (Alpine) |
| DB memory limit | 4 GB |
| `shared_buffers` | 1 GB |
| `work_mem` | 8 MB |
| `maintenance_work_mem` | 256 MB |
| `max_connections` | 200 |
| `shm_size` | 1280 MB |
| Tables per branch schema | ~150 (branchable plugin tables + `django_migrations`) |
| Total tables in `public` | ~394 |
| Total applied migrations | ~862 |
| pgbouncer | Not used — direct connections |
| Branch count at failure | 75+ |

**Expected Behavior:**

`manage.py migrate` completes successfully, marking branches as `PENDING_MIGRATIONS` for subsequent migration via the Migrate action or bulk migrate.

**Observed Behavior:**

All migrations apply successfully to the main schema. The crash occurs afterwards, during the `post_migrate` signal sweep — not while applying DDL or data migrations. The Postgres backend process runs out of memory:

```
psycopg.errors.OutOfMemory: out of memory
DETAIL:  Failed on request of size 284 in memory context "CacheMemoryContext".
```

The traceback bottoms out at:

- `netbox_branching/signal_receivers.py:271` — `check_pending_migrations` iterates every `Branch` with `status=READY`
- `netbox_branching/models/branches.py:458` — `branch.pending_migrations` constructs a `MigrationExecutor` per branch
- `MigrationExecutor.__init__()` calls `introspection.table_names()` which queries `pg_class`

Every `table_names()` call pulls relation metadata into Postgres' session-level `CacheMemoryContext`. The `manage.py migrate` process is one long-lived session, so relcache grows monotonically across all branch iterations and is never released. With ~150 tables per branch schema and 75+ branches, that's 11,000+ relation entries accumulated in a single backend's session memory before any of it is freed.

Note: the signal only fires once (guarded by `sender.name != 'netbox_branching'`), not per-app — but the single invocation iterates all branches in one DB session.

**Root cause:**

`check_pending_migrations` uses `MigrationExecutor` to compute pending migrations per branch. This is heavyweight — it performs full database introspection — but only needs to answer: "are there migrations applied on main that aren't applied on this branch?"

**Suggested fix (any of these would resolve it):**

1. **Close the branch connection between iterations** — call `connections[branch.connection_name].close()` after each `branch.pending_migrations` check. This terminates the Postgres backend, releasing `CacheMemoryContext` before the next iteration. Minimal code change, preserves existing logic.

2. **Precompute main's applied migrations once, then compare per-branch** — query `SELECT app, name FROM public.django_migrations` once at the start of the loop, then for each branch compare against `SELECT app, name FROM <branch_schema>.django_migrations`. Any row present on main but missing on the branch is a pending migration. No `MigrationExecutor`, no introspection, no relcache accumulation. This is O(1) introspection + O(branches) lightweight queries.

3. **Replace `MigrationExecutor` with a single set-difference query per branch** — `SELECT app, name FROM public.django_migrations EXCEPT SELECT app, name FROM <branch_schema>.django_migrations` gives the same answer without any introspection. Variant of (2) with one query per branch instead of in-Python comparison.

4. **Move the check out of the synchronous `post_migrate` signal into a batched RQ job** — this would also solve the wall-clock cost of the detection loop blocking `manage.py migrate`.

We are available to implement whichever approach is preferred, once the issue is accepted.

arthanson:
### Fixes: #581 

The `post_migrate` handler `check_pending_migrations` opens each READY branch's connection to check for pending migrations but never closed them, leaking connections and relcache memory during `manage.py migrate` (where `close_old_branch_connections()` doesn't fire). This closes each branch connection right after inspecting it, with a regression test covering the cleanup.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## v1.0.5
+
+### Bug Fixes
+
+* [#581](https://github.com/netboxlabs/netbox-branching/issues/581) - Close each branch's database connection during the post-migrate pending-migrations check to avoid leaking Postgres backends (and exhausting memory/connections) with many open branches
+
+---
+
 ## v1.0.4
 
 ### Enhancements
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,13 +1,5 @@
 # Change Log
 
-## v1.0.5
-
-### Bug Fixes
-
-* [#581](https://github.com/netboxlabs/netbox-branching/issues/581) - Close each branch's database connection during the post-migrate pending-migrations check to avoid leaking Postgres backends (and exhausting memory/connections) with many open branches
-
----
-
 ## v1.0.4
 
 ### Enhancements
--- a/netbox_branching/signal_receivers.py
+++ b/netbox_branching/signal_receivers.py
@@ -291,6 +291,8 @@ def check_pending_migrations(sender, using, **kwargs):
             if branch.pending_migrations:
                 branch.status = BranchStatusChoices.PENDING_MIGRATIONS
                 update_count += 1
+        except Exception:
+            logger.exception(f"Failed to check pending migrations for branch {branch!r}")
         finally:
             # Close the branch's database connection to release its Postgres backend (and the relcache memory
             # accumulated by the pending_migrations introspection) before moving on. Otherwise these connections
--- a/netbox_branching/signal_receivers.py
+++ b/netbox_branching/signal_receivers.py
@@ -6,7 +6,7 @@ from core.models import ObjectChange, ObjectType
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
-from django.db import DEFAULT_DB_ALIAS
+from django.db import DEFAULT_DB_ALIAS, connections
 from django.db.models.signals import post_migrate, post_save, pre_delete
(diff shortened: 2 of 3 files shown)


arthanson:

arthanson:
pre claude review: addressed 1, second one is minor / not-addressed as doesn't really need a test for that.

jeremystretch (approved):

PR #602, opened 2026-06-18
Add django_apps configuration for PostgreSQL support

ebuildy:
Fix issue:

"django.contrib.postgres" not found in INSTALLED_APPS

when running "manage.py check"

- Netbox v4.4.9
- Netbox Docker 3.4.2

I have no idea why I have to add this line but it fixed our issue.

### DIFF
--- a/netbox_branching/__init__.py
+++ b/netbox_branching/__init__.py
@@ -49,6 +49,7 @@ class AppConfig(PluginConfig):
         # Number of days before staleness at which to display a stale warning
         'stale_warning_threshold': 7,
     }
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thank you for your interest in contributing to NetBox, however it appears there is no accepted issue that correlates to this pull request.  Before pull requests are opened, we require an accepted issue as per our [contributing guide](https://github.com/digitalocean/netbox/blob/main/CONTRIBUTING.md).  Please first open an issue and wait for it to be accepted before further work is done on this pull request.

PR #609, opened 2026-06-29
#608 - Generalize squash merge cycle breaking to FK cycles of any length

### LINKED ISSUE #608: Squash merge fails with "Cycle detected in dependency graph" for multi-node FK cycles (e.g. a device/VM and its primary IP)
### Plugin Version

v1.1.0

### NetBox Version

v4.5.8

### Python Version

3.13

### Steps to Reproduce

1. Create and provision a branch; activate it.
2. Within the branch, create the following objects (all new in the branch):
   1. a `VirtualMachine` (or `Device`)
   2. a `VMInterface` (or `Interface`) on that VM
   3. an `IPAddress`, assigned to that interface (`assigned_object`)
   4. set the VM's `primary_ip4` to that `IPAddress`
3. Merge the branch using the **squash** merge strategy.

This is the ordinary topology for any device/VM that has a primary IP living on one of its own newly-created interfaces.

### Expected Behavior

The merge succeeds. As NetBox's own database does, the VM should be created first with `primary_ip4` left NULL, then the interface, then the IP, and finally the `primary_ip4` FK backfilled once the IP exists.

### Observed Behavior

The squash merge aborts during ordering:

```
Adjusting ordering by references...
  Cycle detected in dependency graph.
    CREATE VirtualMachine (...) depends on: {('ipam.ipaddress', <ip>)}
    CREATE VMInterface (...) depends on: {('virtualization.virtualmachine', <vm>)}
    CREATE IPAddress (<ip>) depends on: {('virtualization.vminterface', <iface>)}
Cycle detected in dependency graph. N changes are involved in circular dependencies and cannot be ordered.
```

The iterative strategy is unaffected (it replays in chronological order and never reorders).

### Root Cause

The collapsed CREATEs form a three-node cycle:

```
VirtualMachine --primary_ip4 (nullable FK)--> IPAddress
IPAddress      --assigned_object (GFK)-------> VMInterface
VMInterface    --virtual_machine (FK)--------> VirtualMachine
```

`SquashMergeStrategy._split_bidirectional_cycles()` in `netbox_branching/merge_strategies/squash.py` only detects and breaks **two-node** (A↔B) cycles: it checks whether a CREATE with a nullable FK to B has B pointing directly back to A. The primary-IP loop is three nodes — the VM's nullable FK (`primary_ip4`) points at the IP, which reaches the VM only transitively via the interface — so it is never broken. `_dependency_order_by_references()` (Kahn's algorithm) then finds no zero-dependency node and raises.

GFK detection (netboxlabs/netbox-branching#467) is a prerequisite for even seeing the loop, since the `IPAddress.assigned_object → VMInterface` edge is generic. This issue is effectively the multi-node generalization of the ordering work in netboxlabs/netbox-branching#448.

### Suggested Fix

Generalize cycle breaking to handle cycles of any length: detect cycles using both concrete FKs and GenericForeignKeys, then break each at an eligible edge backed by a **nullable concrete ForeignKey** by deferring it — set the field NULL on the CREATE and emit a synthetic follow-up UPDATE that restores it once the referenced object exists. This is exactly what the existing two-node splitter does, and it mirrors how NetBox persists `primary_ip4`/`primary_ip6` (and `Circuit.termination_a`) at the database level. The two-node case becomes a length-2 special case of the same routine. Cycles containing no nullable FK to defer should be left for the topological sort to report, as today.

A patch implementing this (generalized cycle breaker plus a regression test that merges a branch containing a VM/Device + interface + IP + `primary_ip4`, including the dual `primary_ip4`/`primary_ip6` variant) is available and can be contributed as a PR.

arthanson:
### Fixes: #608 

Squash merge aborted with "Cycle detected in dependency graph" whenever new objects formed an FK cycle longer than two nodes — most commonly a Device/VM whose primary_ip4 points at an IP on one of its own new interfaces (Device → IPAddress → Interface → Device). The old _split_bidirectional_cycles only detected and broke two-node (A↔B) loops, so this three-node loop was never broken and the topological sort failed.

Replaced _split_bidirectional_cycles (and unused helper _has_fk_to) with a generalized _break_dependency_cycles that handles cycles of any length: it builds a graph over collapsed CREATEs from concrete FKs and GenericForeignKeys, finds cycles via iterative DFS, and breaks each at a nullable concrete FK edge by deferring it (NULL on the CREATE + synthetic follow-up UPDATE) - the same mechanism the old code used, now generalized. 

The two-node case is now the length-2 specialization. Cycles with no deferrable edge are still left for the topological sort to report.

### DIFF
--- a/netbox_branching/merge_strategies/squash.py
+++ b/netbox_branching/merge_strategies/squash.py
@@ -499,6 +499,11 @@ class SquashMergeStrategy(MergeStrategy):
             if color[start] != WHITE:
                 continue
             color[start] = GRAY
+            # Each stack frame stores a *live iterator* over a node's neighbours, not the
+            # neighbour list. Revisiting a frame resumes that iterator where it left off
+            # (Python iterators return self from __iter__), which is what lets DFS continue
+            # past an already-explored child. Do not replace iter(...) with the list itself
+            # or copy the frame's iterator — either would restart the scan and break the DFS.
             stack = [(start, iter(adjacency[start]))]
             while stack:
                 node, neighbors = stack[-1]
@@ -528,7 +533,7 @@ class SquashMergeStrategy(MergeStrategy):
         return None
 
     @staticmethod
-    def _defer_fk(collapsed_changes, create, field_name, logger):
+    def _defer_fk(collapsed_changes, create, field_name):
         """
         Break a cycle at ``create``'s nullable FK ``field_name`` by NULLing it on the CREATE
         and emitting a synthetic follow-up UPDATE that restores the original value once the
@@ -545,13 +550,25 @@ class SquashMergeStrategy(MergeStrategy):
         original_postchange = dict(create.postchange_data)
         create.postchange_data[field_name] = None
 
+        # 3-tuple keys distinguish synthetic UPDATEs from real (2-tuple) changes; assert the
+        # contract so a future 3-tuple key elsewhere can't silently overwrite a real change.
         update_key = (create.key[0], create.key[1], f'update_{field_name}')
+        assert update_key not in collapsed_changes, f"Unexpected key collision: {update_key}"
+
         update_collapsed = CollapsedChange(update_key, create.model_class)
         update_collapsed.change_count = 1  # Synthetic update from split
         update_collapsed.final_action = ActionType.UPDATE
         update_collapsed.prechange_data = dict(create.postchange_data)
         update_collapsed.postchange_data = original_postchange
         update_collapsed.last_change = create.last_change
+
+        # The UPDATE depends on the originating CREATE existing first. That ordering is also
+        # enforced transitively through the cycle (the FK target leads back to this object),
+        # but make it direct so _defer_fk doesn't rely on that implicit precondition.
+        # _build_fk_dependency_graph runs afterwards and only adds to these sets, so the
+        # manual edge survives.
+        update_collapsed.depends_on.add(create.key)
+        create.depended_by.add(update_key)
 
         collapsed_changes[update_key] = update_collapsed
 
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
Claude issues resolved

netbox_branching/merge_strategies/squash.py
```diff
@@ -422,106 +422,250 @@ def _build_fk_dependency_graph(deletes, updates, creates, logger):
...
+    def _iter_create_references(create, creates, ct_cache):
         """
-        Check if a CollapsedChange has a foreign key reference to a specific object.
-        Returns True if any FK or GenericFK field in postchange_data points to the target object.
+        Yield ``(field_name, target_key, breakable)`` for every concrete ForeignKey and
+        GenericForeignKey in ``create.postchange_data`` that points at another object in
+        ``creates`` (self-references excluded). Fields are visited in model definition
+        order, so the yielded sequence is deterministic.
+
+        ``breakable`` is True only for nullable concrete ForeignKeys — the edges a cycle
```
jnovinger, line 432: I don't actually see a `breakable` argument in the method signature.

jnovinger (approved):

netbox_branching/merge_strategies/squash.py
jnovinger, line 432: I don't actually see a breakable argument in the method signature.

jnovinger (changes requested):

netbox_branching/merge_strategies/squash.py
jnovinger: My bad, I now realize it's part of the return value, not a method argument.

jnovinger (approved):

PR #615, opened 2026-07-10
#107 - Auto Archive merged branches

### LINKED ISSUE #107: Add a configuration parameter to automatically archive merged branches
### Plugin Version

v0.4.0

### Proposed functionality

Introduce a new configuration parameter, e.g. `auto_archive_days`, with a default of perhaps 30 days. Branches which were merged more than this long ago will be automatically archived via a daily task.

Setting this parameter to None will disable automatic archiving.

### Use case

This will provide a convenient housekeeping mechanism to automatically clean up old branches which are no longer needed and help combat database bloat.

### External dependencies

_No response_

arthanson:
### Fixes: #107 

Adds a housekeeping mechanism to automatically archive old merged branches, helping combat database bloat from schemas that are no longer needed.

- New auto_archive_days config parameter (default: 30). Merged branches whose merge occurred more than this many days ago are archived automatically. Set to None to disable.
- New AutoArchiveBranchJob that finds MERGED branches past the cutoff and archives them. Configured archive validators are honored (blocked branches are skipped, not force-archived), and per-branch failures are isolated so one bad branch can't abort the batch.
- The plugin's ready() now imports jobs so the system job registers at worker startup.

## Notes

- Archival records a BranchEvent with a null user (system-initiated) and drops the branch's PostgreSQL schema while retaining the Branch record and merged change history - consistent with manual archival.
- Documented under docs/configuration.md.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -23,6 +23,27 @@ A list of import paths to functions which validate whether a branch is permitted
 
 ---
 
+## `auto_archive_days`
+
+Default: `30`
+
+The number of days after which a merged branch is automatically archived. A daily background job archives any branch whose merge occurred more than this many days ago, providing a housekeeping mechanism to clean up old branches which are no longer needed and help combat database bloat.
+
+Set to `None` to disable automatic archival entirely.
+
+```python
+PLUGINS_CONFIG = {
+    'netbox_branching': {
+        'auto_archive_days': 60,
+    }
+}
+```
+
+!!! note
+    Archiving a branch drops its PostgreSQL schema but retains the `Branch` record and its merged change history. Any [archive validators](#archive_validators) configured are honoured; a branch which a validator blocks from being archived is skipped and left in the merged state.
+
+---
+
 ## `exempt_models`
 
 Default: `[]` (empty list)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -25,19 +25,22 @@ A list of import paths to functions which validate whether a branch is permitted
 
 ## `auto_archive_days`
 
-Default: `30`
+Default: `None` (disabled)
 
-The number of days after which a merged branch is automatically archived. A daily background job archives any branch whose merge occurred more than this many days ago, providing a housekeeping mechanism to clean up old branches which are no longer needed and help combat database bloat.
+The number of days after which a merged branch is automatically archived. When set to an integer, a daily background job archives any branch whose merge occurred more than this many days ago, providing a housekeeping mechanism to clean up old branches which are no longer needed and help combat database bloat.
 
-Set to `None` to disable automatic archival entirely.
+Automatic archival is disabled by default. Set this to an integer number of days to enable it.
 
 ```python
 PLUGINS_CONFIG = {
     'netbox_branching': {
-        'auto_archive_days': 60,
+        'auto_archive_days': 30,
     }
 }
 ```
+
+!!! warning
+    Once a branch has been archived its PostgreSQL schema is dropped and it can no longer be reverted. Enable this setting only if you are comfortable with old merged branches becoming non-revertible after the configured number of days.
 
 !!! note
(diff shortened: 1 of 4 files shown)


arthanson:

arthanson:
re claude review - fixes the two issues.

docs/configuration.md
```diff
@@ -23,6 +23,27 @@ A list of import paths to functions which validate whether a branch is permitted
 
 ---
 
+## `auto_archive_days`
+
+Default: `30`
```
jeremystretch: We need to default to disabling this behavior. Otherwise, unsuspecting users may be surprised by branches older than 30 days getting archived soon after upgrading.

netbox_branching/__init__.py
```diff
@@ -86,6 +90,19 @@ def ready(self):
...
+        # than as an opaque timedelta error the first time the daily archival job runs.
+        auto_archive_days = get_plugin_config('netbox_branching', 'auto_archive_days')
+        if auto_archive_days is not None:
+            if type(auto_archive_days) is not int:
+                raise ImproperlyConfigured(
+                    "netbox_branching: 'auto_archive_days' must be an integer or None."
+                )
+            if auto_archive_days < 1:
+                raise ImproperlyConfigured(
+                    "netbox_branching: 'auto_archive_days' must be greater than or equal to 1."
```
jeremystretch:
```suggestion
                    "netbox_branching: 'auto_archive_days' must be greater than or equal to 1 (if enabled)."
```

netbox_branching/jobs.py
```diff
@@ -248,3 +251,57 @@ def run(self, *args, **kwargs):
...
+    class Meta:
+        name = 'Auto-archive branches'
+
+    def run(self, *args, **kwargs):
+        # Import here to avoid a circular import at module load time (models imports jobs).
+        from .models import Branch
+
+        auto_archive_days = get_plugin_config('netbox_branching', 'auto_archive_days')
+        if auto_archive_days is None:
+            self.logger.info("Automatic branch archival is disabled (auto_archive_days is None); skipping.")
```
jeremystretch: IMO we should probably not log a message if auto-archiving is disabled to avoid noise. Or at least change it from info to debug.

netbox_branching/jobs.py
```diff
@@ -248,3 +251,57 @@ def run(self, *args, **kwargs):
...
+        )
+        self.logger.info(
+            f"Found {len(branch_pks)} merged branch(es) merged before {cutoff:%Y-%m-%d %H:%M:%S} "
+            f"eligible for automatic archival."
+        )
+
+        for pk in branch_pks:
+            # Re-fetch each branch for its current state; it may have been archived or deleted
+            # since the initial query was taken.
+            branch = Branch.objects.filter(pk=pk).first()
```
jeremystretch: Can we just pull all the branches initially, instead of querying for PKs first and then retrieving them individually? I assume the goal here is to mitigate a race condition, but it seems pretty unlikely given we're specifically targeting dormant branches. Your call though.

docs/configuration.md
```diff
@@ -23,6 +23,27 @@ A list of import paths to functions which validate whether a branch is permitted
...
+```python
+PLUGINS_CONFIG = {
+    'netbox_branching': {
+        'auto_archive_days': 60,
+    }
+}
+```
+
+!!! note
+    Archiving a branch drops its PostgreSQL schema but retains the `Branch` record and its merged change history. Any [archive validators](#archive_validators) configured are honoured; a branch which a validator blocks from being archived is skipped and left in the merged state.
```
jeremystretch, line 46: Maybe just reiterate explicitly that archived branches cannot be reverted.

jeremystretch (changes requested):
The only real concern I have is enabling auto-archival by default. Let's keep it opt-in to avoid any nasty surprises, _especially_ in a patch release.

jeremystretch (approved):
Thanks @arthanson!

PR #620, opened 2026-07-20
#617 - Support Changes for NetBox 4.7 LTree

### LINKED ISSUE #617: Support Changes for NetBox 4.7 MPTT / LTree Changes
### Plugin Version

v1.1.1

### Proposed functionality

Make changes in the branching plugin to support the changes in NetBox 4.7 switch form MPTT to LTree.

### Use case

Braching has a couple places where it needs to check if a model is derived from MPTT for special handling, these will need to be updated because LTree has different behavior then MPTT that will effect branching.

### External dependencies

N/A

arthanson:
### Fixes: #617 

NetBox 4.7 replaces django-mptt with a PostgreSQL `ltree` implementation for hierarchical models, and moves both ltree path maintenance and the device-component `_site`/`_location`/`_rack` denormalization from Python into **per-table database triggers**. This PR makes branching work correctly against 4.7.

**Note:** Targets feature branch in CI as that has the changes needed for this to work (these changes are post Beta1)

### The core problem

Branch provisioning creates schema tables with `CREATE TABLE ... (LIKE ...)`, and PostgreSQL's `LIKE` **does not copy triggers**. MPTT and the old denormalization ran in Python, so they worked inside branches. 

Now that this logic lives in triggers, branch schemas were missing it - e.g. reparenting a Region inside a branch left `path`/`level` stale.

### Changes

- **Replicate triggers on provision** (`models/branches.py`): a new phase copies every non-internal trigger from main's tables onto the branch copies. Trigger functions are schema-agnostic (resolve their table via `search_path`), so branch triggers reuse the existing `public` functions while operating on branch data. Covers both ltree and denormalization triggers.
- **Fix M2M loss on revert** (`models/changes.py`): restoring a deleted object now goes through `DeserializedObject.save()`, re-establishing M2M relations (e.g. tags on a reverted `ModuleBay`). Previously only the MPTT path did this.
- **Merge strategy** (`merge_strategies/strategy.py`): ltree models need no post-merge rebuild (triggers keep paths consistent); MPTT `rebuild()` retained for the still-shipped `NestedGroupModel`.
- **Tests**: ported tree assertions from `tree_id`/`lft`/`rght` to `path`.
- **Compatibility**: requires NetBox 4.7 (`min_version`/`max_version` → 4.7.x), bump to v1.2.0, matrix updated.


arthanson:

arthanson:
Per claude review:

1. Advisory lock deadlock - not fixed (invalid as described). Upstream takes both locks in ascending key order (LEAST/GREATEST at utilities/ltree.py:69-70), so the opposite-order deadlock can't occur; the real effect is only redundant branch-vs-main lock contention..
2. EXEMPT_MODEL_FIELDS ltree fields - not fixed (not applicable). The test only walks Branch/BranchEvent/ChangeDiff fields, so no hierarchical model is ever covered and path/sort_path would be dead config.
3. CI NetBox ref - temporary by design so CI will actually work
4. mutate_state(preserve=False) - not fixed (invalid). Django uses preserve=False the same way in four places; the code keeps no reference to the  pre-mutation state, so nothing can be corrupted.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "netboxlabs-netbox-branching"
-version = "1.1.2"
+version = "1.2.0"
```
jnovinger, line 3: I think we generally leave this until the release PR, don't we?

jnovinger (approved):

PR #625, opened 2026-08-12
Fixes #624: Ensure conflicts are recorded in ChangeDiff records during object updates

### LINKED ISSUE #624: Sync/merge fails with a `face` ValidationError when rack/position/face diverge between a branch and main
### Plugin Version

1.1.2

### NetBox Version

4.6.8

### Python Version

3.14

### Steps to Reproduce

1. Create the prerequisite objects: a manufacturer, device type, device role, site, and a rack (`rack01`) at that site.
2. Create a device `router01` in a racked, front-mounted baseline state: `rack=rack01`, `position=1`, `face=front`.
3. Create a branch (e.g. `sync-conflict-S`) and wait for it to finish provisioning.
4. **In the branch**, unrack the device in a single update: `rack=null`, `position=null`, `face=""`. This is a self-consistent change and is accepted.
5. **In main**, independently change only the device's rack face: `face=front` → `face=rear`. Also self-consistent and accepted, since `rack` and `position` are still set in main.
6. Sync the branch: `POST /api/plugins/branching/branches/{id}/sync/`, then poll the resulting job.

The mirrored sequence reproduces the same failure on merge instead of sync: make the `face`-only change in the branch, unrack the device in main, then merge the branch.

### Expected Behavior

The sync completes. Main's `face`-only change is replayed against a branch whose device is already unracked, so the result should either preserve the branch's unracked state (the branch's change is the more recent one from the branch's perspective) or be surfaced as a conflict that names all of the mutually dependent fields involved (`rack`, `position`, `face`) so the user knows what actually has to be reconciled.

At minimum, the sync should not become permanently unrunnable due to two changes that are each individually valid.

### Observed Behavior

The sync job errors out and the branch can never be synced until the fields are reconciled by hand in the branch or in main:

```
sync job status: errored
sync job error:  ValidationError({'face': ['Cannot select a rack face without assigning a rack.']})
  Syncing branch sync-conflict-S (branch_ucxy7rf2)
  Found 1 changes to sync
  Setting branch status to syncing
  Applying change 45 using schema_branch_ucxy7rf2 (DCIM | device router01 updated by nbl)
  Updating device router01
  {'face': ['Cannot select a rack face without assigning a rack.']}
```

The branch is not corrupted — the sync runs inside `transaction.atomic()`, so it rolls back and the branch returns to `ready` — but the sync fails identically on every retry.

### Analysis

`Device.clean()` requires `rack`, `position`, and `face` to be mutually consistent:

```python
if self.rack is None:
    if self.face:
        raise ValidationError({'face': "Cannot select a rack face without assigning a rack."})
    if self.position:
        raise ValidationError({'position': "Cannot select a rack position without assigning a rack."})
if self.position and not self.face:
    raise ValidationError({'face': "Must specify rack face when defining rack position."})
```

When a change is replayed during sync/merge, `update_object()` in `netbox_branching/utilities.py` `setattr()`s only the fields present in that single `ObjectChange` diff, then calls `full_clean()` on the *entire* object:

```python
def update_object(instance, data, using):
    ...
    for attr, value in data.items():
        ...
        setattr(instance, attr, value)

    full_clean_with_file_check(instance, logger)   # validates ALL fields, not just those in `data`
    instance.save(using=using)
```

So the validity of a replayed change depends on fields the change never touched. Replaying main's diff (`{'face': 'rear'}`) onto the branch's current state (`rack=None`) produces an object that fails `full_clean()`, even though neither side's change is invalid on its own.

The reported field is also misleading: the error names `face` (or whichever field happens to be in the replayed diff), while the actual missing piece is `rack`.

jeremystretch:
### Fixes: #624

Records `ChangeDiff` conflicts when the change in main arrives *after* the branch's change, so a divergence like the one in #624 is surfaced as a conflict the user must acknowledge instead of a sync/merge job that errors out with no prior warning.

## The bug

`record_change_diff()` handles a change made in main by refreshing `ChangeDiff.current` for every branch that has touched the same object. It did so with a queryset `update()`:

```python
ChangeDiff.objects.filter(...).update(
    last_updated=timezone.now(),
    current=instance.postchange_data_clean or None
)
```

A queryset `update()` bypasses `Model.save()`, and `_update_conflicts()` only runs from `ChangeDiff.save()`. So `current` was refreshed while `conflicts` was left at whatever the branch's own change had computed earlier — normally `NULL`. A main-side change could therefore introduce a genuine field-level conflict that was never recorded.

Everything that reads the field was affected: the API's 409 conflict gate (`BranchViewSet._check_conflicts`), the merge form's "all conflicts must be acknowledged" check, the conflict indicators in the diff table, and the `has_conflicts` filter.

The ordering is what makes this easy to miss. When main changes *first*, the branch's subsequent change re-saves the `ChangeDiff` and recomputes conflicts as a side effect — and every existing conflict test in the suite happens to use that order. Only main-after-branch hits the broken path, which is exactly the sequence in #624 (the branch unracks the device, then main changes `face`).

The main-side DELETE case already needed its own recompute loop to work around the bulk update; that workaround is now unnecessary.

## The fix

Save each `ChangeDiff` individually so `_update_conflicts()` re-runs:

```python
for diff in ChangeDiff.objects.filter(...).select_related('object_type'):
    diff.current = current_data
    try:
        with transaction.atomic():
            diff.save(update_fields=('current', 'conflicts', 'last_updated'))
    except DatabaseError:
        logger.debug("ChangeDiff %s no longer exists; skipping update", diff.pk)
```

`update_fields` deliberately omits `object_repr`. `save()` recomputes it from the GenericForeignKey, which resolves against main here (there is no active branch in this path), and the stored value must keep reflecting the branch's view of the object — otherwise renaming an object in a branch and then editing it in main would relabel the branch's diff row with main's name.

The DELETE-specific recompute loop is removed, as the unified loop covers it.

### `ChangeDiff.save()` now honours `update_fields`

```python
update_fields = kwargs.get('update_fields')
if (update_fields is None or 'object_repr' in update_fields) and self.object:
    self.object_repr = str(self.object)
```

`save()` previously resolved the GenericForeignKey and called `str()` on the result unconditionally, then discarded the value whenever `object_repr` was excluded from `update_fields` — as it is in the loop above. That is a wasted `SELECT` per diff, plus whatever the model's `__str__` costs (an unnamed `Device.__str__` walks `device_type.manufacturer`, so 2–3 queries).

It matters because this is a hot path during a merge: `IterativeMergeStrategy.merge()` writes a main-side `ObjectChange` per applied change inside `event_tracking()`, so the loop runs once per applied change for every *other* READY branch holding a diff on that object. The waste scales with changes × branches, inside the merge transaction. Full saves (`update_fields=None`) still recompute `object_repr`, so behaviour is unchanged.

### Zero-row saves

`save(update_fields=...)` raises when the row matches nothing, where the previous bulk `update()` was silent on zero matches. `ChangeDiff.branch` is `on_delete=CASCADE`, so a branch deleted or deprovisioned between this loop's `SELECT` and a given row's `save()` reaches that path. Two details govern how it has to be handled:

**The exception type depends on the Django version.** Django 6.0 raises `Model.NotUpdated`, but Django 5.2 — which NetBox 4.4.1 (`Django==5.2.6`) and 4.5.0 (`Django==5.2.9`) pin, against a declared `min_version` of 4.4.1 — raises a plain `DatabaseError`, and `ChangeDiff.NotUpdated` does not exist there at all. Naming it in the `except` clause would raise `AttributeError` on the two oldest supported NetBox versions, replacing the original error with a more confusing one. Django 6.0 deliberately makes the per-model exception a `DatabaseError` subclass for backward compatibility (`NotUpdated → ObjectNotUpdated → DatabaseError`), so a single `except DatabaseError` covers NetBox 4.4 through 4.6.

**Catching it is not sufficient on its own.** `Model.save_base()` wraps `_save_table()` in `transaction.mark_for_rollback_on_error()`, which sets `needs_rollback` on the connection whenever the save raises inside an atomic block — and this receiver runs inside the view's transaction. Swallowing the exception without a savepoint therefore leaves the connection poisoned and the unrelated write to main still fails, just as `TransactionManagementError` rather than the original error. The inner `transaction.atomic()` gives the failed save a savepoint to roll back to, so the enclosing transaction stays usable.

### Divergent key sets in `_update_conflicts()`

```python
conflicts = [k for k, v in original.items() if v != modified.get(k)]
```

`original` is captured from the branch's first change while `modified` is refreshed on every subsequent one, so their key sets can diverge across a NetBox upgrade that adds or renames a field with no registered migrator. The previous `modified[k]` raised `KeyError` on such a key. That line used to be reachable only from a branch-side save; now that a main-side change recomputes conflicts, the exception would propagate out of the `post_save` receiver and roll back an unrelated write to main.

`modified.get(k)` matches the `current.get(k)` already used alongside it. A key missing from `modified` compares as `None` and is reported as a conflict, which surfaces the divergence for the user to resolve rather than silently merging past it.

### Query cost

An object with no branch diffs now costs one `SELECT` returning zero rows instead of one `UPDATE` matching zero rows. Where diffs do exist, there is at most one per READY branch that has touched that object, and each is now a bare `UPDATE` with no GFK resolution behind it.

The debug log at the top of the same path is guarded with `logger.isEnabledFor(logging.DEBUG)`. `instance.changed_object` is a GenericForeignKey resolved eagerly as a call argument, so `logger.debug("... %s", instance.changed_object)` defers the *formatting* but not the `SELECT` — it would otherwise cost a query on every main-side change regardless of log level, on the same path where skipping `object_repr` saves one.

## Tests

New `MainSideConflictTestCase` in `tests/test_changediff.py`, going through the real signal path with a provisioned branch:

- a conflict is recorded when main changes the same field after the branch did;
- no false conflict when main changes a field the branch never touched;
- an existing conflict is *cleared* when a later change in main converges on the branch's value;
- `object_repr` still reflects the branch after a main-side change;
- the branch-UPDATE + main-DELETE combination still records a conflict (coverage for the removed loop);
- a `ChangeDiff` deleted between the loop's `SELECT` and its `save()` is skipped without aborting the enclosing write to main.

New `DivergentKeySetTestCase` covers `_update_conflicts()` when a key in `original` is absent from `modified`, both with and without the object still present in main.

Each of these fails with its corresponding fix reverted:

| Test | Failure without the fix |
|---|---|
| `test_conflict_recorded_for_main_change_after_branch_change` | `None != ['description']` |
| `test_conflict_cleared_when_main_matches_branch_value` | `None != ['description']` |
| `test_deleted_diff_does_not_abort_main_write` | `TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.` |
| `DivergentKeySetTestCase` (both) | `KeyError: 'retired'` |

The remainder are guards against regressing the surrounding behaviour.

## Verification

- `netbox_branching.tests.test_changediff` — 29 tests, OK, against a NetBox `main` checkout on Django 6.0.8.
- Every new regression test confirmed to fail with its fix reverted, as tabulated above.
- `ruff check` clean.

Two caveats on coverage:

- The Django 5.2 behaviour described under "Zero-row saves" is established from the Django sources (`Model.NotUpdated` was introduced by `ab148c02ce`, first released in 6.0; `git show 5.2:django/db/models/base.py` has no reference to it), not by execution — only Django 6.0.8 is installed locally. CI's NetBox matrix is what will exercise the 5.2 path.
- The wider comparison run (`test_iterative_merge`, `test_squash_merge`, `test_sync`, `test_api`, `test_views` — 231 tests, run twice against the same database, once with these changes and once with them stashed) was performed against the first version of this fix and has **not** been repeated since the changes above. It reported 1 failure / 1 error / 2 skipped identically in both runs; both are pre-existing against a NetBox `main` checkout:
  - `test_merge_after_sync_cascade_delete_of_branch_modified_object` raises `select_for_update cannot be used outside of a transaction` from inside NetBox's own `dcim.signals.handle_location_site_change`;
  - `test_list_objects_with_permission` trips a stale query-count baseline (`query_counts.json` expects 21; NetBox `main` now issues 19). None of the 19 observed queries touch this code path.

## Not addressed here

This makes the #624 divergence *visible and gated* rather than a job failure with no warning, but it does not make that sync succeed. The underlying cause of the `ValidationError` is separate: `update_object()` replays only the fields in a single `ObjectChange` diff and then calls `full_clean()` on the whole object, so a replay's validity depends on fields it never touched, and `Device.clean()` enforces a relationship between `rack`, `position`, and `face`.

Because conflict detection is per-field, it also cannot express that class of divergence. There are combinations with **no** field-level conflict at all that still fail — e.g. baseline `rack=rack01, position=null, face=front`, the branch sets `position=2`, main clears `face`: each change is valid alone, neither side touches the same field, and replaying main's `{'face': ''}` onto the branch raises `{'face': ['Must specify rack face when defining rack position.']}`.

Fixing that means handling the `ValidationError` during replay — recording it against the object's `ChangeDiff` as an unresolvable conflict naming all the mutually dependent fields, instead of aborting the job — and is left for a follow-up.

### DIFF
--- a/netbox_branching/models/changes.py
+++ b/netbox_branching/models/changes.py
@@ -251,7 +251,10 @@ class ChangeDiff(models.Model):
 
     def save(self, *args, **kwargs):
         self._update_conflicts()
-        if self.object:
+        # Resolving the object (a GenericForeignKey) and rendering it as a string can require several
+        # queries, so skip it entirely when object_repr is not among the fields being saved.
+        update_fields = kwargs.get('update_fields')
+        if (update_fields is None or 'object_repr' in update_fields) and self.object:
             self.object_repr = str(self.object)
 
         super().save(*args, **kwargs)
@@ -281,11 +284,11 @@ class ChangeDiff(models.Model):
         if self.action == ObjectChangeActionChoices.ACTION_UPDATE:
             if current is None:
                 # Object was deleted in main; all branch modifications are in conflict
-                conflicts = [k for k, v in original.items() if v != modified[k]]
+                conflicts = [k for k, v in original.items() if v != modified.get(k)]
             else:
                 conflicts = [
                     k for k, v in original.items()
-                    if v != modified[k] and v != current.get(k) and modified[k] != current.get(k)
+                    if v != modified.get(k) and v != current.get(k) and modified.get(k) != current.get(k)
                 ]
         elif self.action == ObjectChangeActionChoices.ACTION_DELETE:
             if current is None:
--- a/netbox_branching/signal_receivers.py
+++ b/netbox_branching/signal_receivers.py
@@ -6,10 +6,9 @@ from core.models import ObjectChange, ObjectType
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
-from django.db import DEFAULT_DB_ALIAS, connections
+from django.db import DEFAULT_DB_ALIAS, DatabaseError, connections, transaction
 from django.db.models.signals import post_migrate, post_save, pre_delete
 from django.dispatch import receiver
-from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from extras.events import process_event_rules
 from extras.models import EventRule
@@ -134,26 +133,26 @@ def record_change_diff(instance, **kwargs):
         if instance.action == ObjectChangeActionChoices.ACTION_CREATE:
             return
 
-        logger.debug(f"Updating change diff for global change to {instance.changed_object}")
-        ChangeDiff.objects.filter(
+        if logger.isEnabledFor(logging.DEBUG):
+            # changed_object is a GenericForeignKey, so it is resolved eagerly as an argument even
+            # though logger.debug() would defer the formatting.
+            logger.debug("Updating change diff for global change to %s", instance.changed_object)
+        current_data = instance.postchange_data_clean or None
+        for diff in ChangeDiff.objects.filter(
             object_type=content_type,
             object_id=object_id,
             branch__status=BranchStatusChoices.READY
-        ).update(
-            last_updated=timezone.now(),
-            current=instance.postchange_data_clean or None
-        )
-
-        # When main deletes an object, recompute conflicts for any branch edits on that object,
-        # as the branch-UPDATE + main-DELETE combination is not caught by the bulk update above.
-        if instance.action == ObjectChangeActionChoices.ACTION_DELETE:
-            for diff in ChangeDiff.objects.filter(
-                object_type=content_type,
-                object_id=object_id,
-                branch__status=BranchStatusChoices.READY,
-                action=ObjectChangeActionChoices.ACTION_UPDATE,
-            ):
-                diff.save()
+        ).select_related('object_type'):
+            diff.current = current_data
+            # object_repr is excluded from update_fields because save() recomputes it from the
+            # GenericForeignKey, which resolves against main here (no active branch). The stored
+            # value must keep reflecting the branch's view of the object.
+            try:
+                with transaction.atomic():
+                    diff.save(update_fields=('current', 'conflicts', 'last_updated'))
+            except DatabaseError:
+                # The ChangeDiff was deleted (e.g. along with its Branch) after being retrieved above.
+                logger.debug("ChangeDiff %s no longer exists; skipping update", diff.pk)
 
     # If this is a branch-aware change, create or update ChangeDiff for this object.
     else:
--- a/netbox_branching/tests/test_changediff.py
+++ b/netbox_branching/tests/test_changediff.py
@@ -1,10 +1,21 @@
+import uuid
 from datetime import timedelta
 
 from core.choices import ObjectChangeActionChoices
+from dcim.models import Site
+from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
-from django.test import SimpleTestCase, TestCase
+from django.db import connections, transaction
+from django.db.models.signals import post_save
+from django.test import RequestFactory, SimpleTestCase, TestCase, TransactionTestCase
+from django.urls import reverse
+from netbox.context_managers import event_tracking
 
 from netbox_branching.models import Branch, ChangeDiff
+from netbox_branching.tests.utils import provision_branch
+from netbox_branching.utilities import activate_branch
+
+User = get_user_model()
 
 DATA_A = {'name': 'foo', 'description': ''}
 DATA_B = {'name': 'foo', 'description': 'changed'}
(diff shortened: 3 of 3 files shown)


jeremystretch:

jeremystretch:
Opted to skip the above to preserve compatibility with older releases. Once we drop support for NetBox older than v4.6, we can adopt `Model.NotUpdated`.

jnovinger (approved):

PR #635, opened 2026-08-26
#634 publish pre-release builds to PyPi from tag pushes

### LINKED ISSUE #634: Publish pre-release builds to PyPI from tag pushes
### Proposed Changes

Switch release.yaml from triggering on release: published to triggering on a v* tag push, accepting tags of the form vX.Y.Z[-designation] so v1.3.0-beta1 publishes as 1.3.0b1. 

### Justification

We need to ship betas of the plugin. The current workflow keys off a published GitHub release and has no version verification, which means a beta can only go out by hand

This mirrors what NetBox itself moved to in its own release workflow.

arthanson:
### Fixes: #634

Switch release.yaml from triggering on release: published to triggering on a v* tag push, accepting tags of the form vX.Y.Z[-designation] so v1.3.0-beta1 publishes as 1.3.0b1. Add packaging verification ahead of the upload, plus a wheel-contents check that the templates and migrations shipped

After PyPI accepts the upload, create the GitHub release if the tag doesn't already have one, marking any designated tag as a pre-release. Add an opt-in Test PyPI route via manual dispatch for rehearsing a publish.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,9 +1,24 @@
 name: Release
 
+# Releases are driven by pushing a v* tag, not by publishing a GitHub release, so that
+# pre-releases (v1.3.0-beta1 -> 1.3.0b1 on PyPI) follow the same automated path as a
+# final release. The GitHub release is created by this workflow after PyPI accepts the
+# upload; see the release job below.
 on:
-  release:
-    types: [published]
+  push:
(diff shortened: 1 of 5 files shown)


arthanson:

arthanson:
Claude issues addressed.

jnovinger (approved):

PR #644, opened 2026-08-31
#622 Recover branches stuck in a transitional status when a worker dies

### LINKED ISSUE #622: Branch "stuck" in Migrating when worker fails
### Plugin Version

1.1.1

### NetBox Version

4.6.5

### Python Version

3.13

### Steps to Reproduce

Reporting this as it caused us a lot of headache. Root cause is under-resourced worker, but perhaps could be handled better by the plugin: 

1. Create a branch on an earlier Netbox version (with substantial data)
2. Upgrade Netbox
3. Perform migrations on the branch
4. Kill the worker performing the job while it is in progress

### Expected Behavior

* Job times out if worker is lost / time out in some way -- this may be a bug / limitation of Netbox jobs as opposed to Netbox branching specifically 

* Branch reverts to "pending migrations" -- ideally if there is no progress on the job or the job has failed, the branch should revert so that the user can re-attempt the migrations. 

### Observed Behavior

When a worker fails to complete a migration job (in our case, it would appear Kubernetes was under-resourced to complete the migration job and was killed off), the job remains stale as "Running" and the Branch is stuck in a "Migrating" state. The UI is greyed out and you cannot stop or reset the state to re-attempt the migration. There is no feedback in the UI for what happened. 

Only manually setting the branch back to "pending-migration" 

For others who may encounter this issue, this is how I resolved
```
BEGIN;
UPDATE public.netbox_branching_branch
SET
    status = 'pending-migrations',
    last_updated = now()
WHERE
   AND status = 'migrating';
COMMIT;

### PR DESCRIPTION
**Pull request by arthanson**
### Fixes: #622 

 A failed branch operation can leave a branch stuck in a `provisioning`, `syncing`, `migrating`, `merging` or `reverting` state - when a worker is killed outright it never gets to clear the status, so every action stays disabled and a manual `UPDATE` is the only way out.

This PR adds:

* An hourly job that resets these unattended.
* A **Recover** button (and `recover/` REST endpoint) that lets the user reset the branch and, for a stuck sync or migrate, re-run the operation immediately.

A stuck migrate has the reset button:
Monosnap 622 stuck migrating | NetBox 2026-09-03 15-26-01

And here is the reset migration dialog:
Monosnap Recover Branch 622 stuck migrating | NetBox 2026-09-03 15-26-22

Sync also has a reset button:
Monosnap 622 stuck syncing | NetBox 2026-09-03 15-26-38

and redo sync dialog:
Monosnap Recover Branch 622 stuck syncing | NetBox 2026-09-03 15-26-52

### CODE REVIEW

**Comment by arthanson** [comment]

**Comment by arthanson** [comment]
All three Claude review comments addressed and fixed.

File: netbox_branching/utilities.py [hunk 1]
```diff
@@ -627,3 +640,62 @@ class BranchActionIndicator:
...
+    Return the RQ status of the RQ job backing a NetBox `Job`, the string ``'missing'`` if RQ no
+    longer knows about it, or None if RQ could not be consulted (e.g. Redis is unreachable).
+    """
+    import django_rq
+    from utilities.rqworker import get_queue_for_model
+
+    try:
+        # Fall back to the queue for the job's object type, matching Job.delete()'s handling of
+        # legacy jobs recorded before queue_name was introduced.
+        queue_name = job.queue_name or get_queue_for_model(job.object_type.model if job.object_type else None)
```
jnovinger, line 660: My local Claude caught that `job.queue_name` was introduced in or around NetBox v4.5.2, but the plugin claims to support down to v4.4.1. The upshot, I think, is that this generates an `AttributeError` in 4.4.x and 4.5.0/4.5.1 (anything below v4.5.2) which gets immediately swallowed.

I'm not sure I'd go so far as to say this break compatibility with <v4.5.2, but it certainly seems to render this fix inoperative on those versions.

The tests won't catch this either, since they all patch `_get_rq_job_status` itself and never execute this line.

Was 4.4.x support meant to be dropped here, or should this fall back the way `Job.delete()` does on that version?

netbox_branching/utilities.py
```diff
@@ -627,3 +640,62 @@ class BranchActionIndicator:
...
         return self.permitted
+
+
+def _get_rq_job_status(job):
+    """
+    Return the RQ status of the RQ job backing a NetBox `Job`, the string ``'missing'`` if RQ no
+    longer knows about it, or None if RQ could not be consulted (e.g. Redis is unreachable).
+    """
+    import django_rq
+    from utilities.rqworker import get_queue_for_model
```
jnovinger, line 655: Do theses imports need to be nested? My gut says `django_rq` does not, but unsure about the other.

netbox_branching/choices.py
```diff
@@ -61,6 +61,53 @@ class BranchStatusChoices(ChoiceSet):
         *TRANSITIONAL,
     )
 
+    # The status to which a branch is reset when the background job responsible for a
+    # transitional status is no longer running (e.g. its worker was killed). Each value
+    # mirrors the status that the corresponding branch operation restores itself when it
```
jnovinger: > Each value mirrors the status that the corresponding branch operation restores itself

This says each value mirrors what the operation restores on failure, but migrate's failure path sets `FAILED` (`branches.py:945`, deliberately, so a partially-migrated branch can't be activated), not `PENDING_MIGRATIONS`. The mapping itself looks right, it's just the blanket claim up here that doesn't seem to hold.

netbox_branching/api/views.py
```diff
@@ -168,6 +169,41 @@ def archive(self, request, pk):
...
+
+        branch = self.get_object()
+        if branch.status not in BranchStatusChoices.TRANSITIONAL:
+            return HttpResponseBadRequest("Branch is not in a transitional status.")
+
+        # Recover only a branch which is demonstrably stuck, unless the caller explicitly forces it.
+        serializer = serializers.BranchRecoverSerializer(data=request.data)
+        params = serializer.validated_data if serializer.is_valid() else {}
+        # Re-running the interrupted operation is opt-in over the API, where there is no confirmation
+        # step; the UI offers it as a checked-by-default option instead.
```
jnovinger: Is this comment accurate? `forms/misc.py:145` says `initial=False`.

netbox_branching/__init__.py
```diff
@@ -59,6 +59,14 @@ class AppConfig(PluginConfig):
         # Automatically archive branches merged more than this many days ago (via a daily job).
         # Set to an integer number of days to enable automatic archival.
         'auto_archive_days': None,
+
+        # Automatically reset branches left in a transitional status (e.g. "migrating") by a
+        # background job whose worker died before it could complete. Checked hourly.
+        'auto_recover_stuck_branches': True,
```
jnovinger, line 65: Given that this is targeting `main`, this will go out in a patch release. Is it appropriate to set the default behavior to `True` in a patch release? Not disagreeing necessarily, just felt odd.

jnovinger (changes requested):

netbox_branching/utilities.py
arthanson: queue_name does land in 4.5.2, but 1.2.0 already raised min_version to 4.7.0 — the stale 4.4.1 claim was in AGENTS.md, now fixed. test gap fixed.

netbox_branching/utilities.py
arthanson: they need to be nested

netbox_branching/__init__.py
arthanson: I get it, but I'd sort of consider it on the side of a bug fix - if a branch was stuck before there was no way to get it un-stuck (so they become un-usable), so this is just getting them un-stuck and usable.

netbox_branching/utilities.py
jnovinger: My apologies, I got confused here.

jnovinger (approved):

PR #647, opened 2026-09-02
Release v1.2.0

arthanson:
### NetBox v4.7 Support ([#617](https://github.com/netboxlabs/netbox-branching/issues/617))

This release adds support for NetBox v4.7, which replaces django-mptt with a PostgreSQL ltree implementation and moves the maintenance of ltree and denormalized fields into database triggers. Branch provisioning now replicates those triggers into each branch schema, so nested object hierarchies behave consistently within a branch.

**NetBox v4.7 is now required.** Earlier NetBox releases are not supported by this version; continue to use the 1.1.x series with NetBox v4.4 through v4.6.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "netboxlabs-netbox-branching"
-version = "1.2.0-beta1"
+version = "1.2.0"
 description = "A git-like branching implementation for NetBox"
 readme = "README.md"
 requires-python = ">=3.10"
```
pheus: **Blocking:** Could we align the Python metadata with the NetBox 4.7 requirement as part of this release? NetBox 4.7 requires Python 3.12+, but `requires-python` still permits Python 3.10/3.11 and the classifiers stop at Python 3.12, while CI covers Python 3.12–3.14.

Please update `requires-python` to `>=3.12`, remove the Python 3.10/3.11 classifiers, and add Python 3.13/3.14. Otherwise PyPI will advertise v1.2.0 for runtimes on which its only supported NetBox series cannot run.

pheus (changes requested):

pheus (approved):

PR #648, opened 2026-09-03
Fixes: #650 - Fix ObjectChange.undo() bypassing model save() on non-MPTT restore

### LINKED ISSUE #650: ObjectChange.undo() bypasses model save() on non-MPTT restore, breaking auto_now fields and save() overrides
### Plugin Version

1.2.0

### NetBox Version

4.7.0

### Python Version

3.12

### Steps to Reproduce

1. Create a plain (non-MPTT) model instance with a many-to-many relationship, e.g. a `Site` with tags:
   ```python
   site = Site.objects.create(name='Site 1', slug='site-1')
   site.tags.add(tag1, tag2)
   ```
2. Create and provision a branch, then delete the object inside it:
   ```python
   branch = provision_branch(...)
   with activate_branch(branch):
       Site.objects.get(id=site.id).delete()
   ```
3. Merge the branch (the object is now deleted in main).
4. Revert the branch: `branch.revert(user=..., commit=True)`.

A more production-shaped reproduction, using a model that (a) has a `save()` override with a real side effect, and (b) excludes an `auto_now` field from `serialize_object()` to keep it out of change-log diffs (a pattern NetBox core itself uses for `ConfigContext`'s cache fields), is exercised by `netboxlabs-netbox-custom-objects`' own test suite:
- `netbox_custom_objects.tests.test_branching.IterativeBranchingTestCase.test_cot_deleted_in_branch_merge_and_revert`
- `netbox_custom_objects.tests.test_branching.IterativeBranchingTestCase.test_single_field_delete_merge_and_revert`

Both fail against netbox-branching 1.2.0 and pass against 1.1.2 with no other changes.

### Expected Behavior

Restoring a deleted object during `branch.revert()` should behave like any other real save of that model: any `auto_now`/`auto_now_add` field should be populated normally, and any `save()` override the model defines should run (including its side effects), the same way it always has for every other code path that creates or restores an object.

### Observed Behavior

Restoring a deleted object goes through `DeserializedObject.save()`, which calls `Model.save_base(..., raw=True)` directly. Per Django's own source comment, this "bypasses any model-defined save" and also skips `Field.pre_save()`, which is what normally populates `auto_now`/`auto_now_add` fields on a real save. Two concrete failure modes result:

- A model that excludes an `auto_now` field from `serialize_object()` (to keep it out of change-log diffs) gets that field restored as whatever value the deserializer defaulted it to for a field missing from the payload -- typically `None` -- which raises a NOT NULL violation if the column is non-nullable. (`netbox_custom_objects.CustomObjectType.cache_timestamp`.)
- A model's `save()` override that performs necessary side effects on creation (e.g. applying schema DDL) never runs, so the object's own DB row comes back but the side effect doesn't happen. (`netbox_custom_objects.CustomObjectTypeField.save()`, which adds the physical column for a new field -- reverting a field deletion restores the field's metadata row but not its actual table column.)

### Suspected Cause

`ObjectChange.undo()`'s non-MPTT branch (`netbox_branching/models/changes.py`) calls `deserialized.save(using=using)`, where `deserialized` is the `DeserializedObject` wrapper returned by `deserialize_object()` -- not the underlying model instance. This was changed from `instance.save(using=using)` (the real model instance's own `.save()`) in #647 (v1.2.0), specifically to fix M2M data loss on restore for non-MPTT models. The MPTT branch immediately above it already demonstrates the pattern that avoids this trade-off: call the real instance's `.save()`, then explicitly reapply `deserialized.m2m_data` afterward. The non-MPTT branch could do the same, preserving the M2M-preservation fix without bypassing `auto_now` population or `save()` overrides.

### Proposed Fix

In the non-MPTT branch of `ObjectChange.undo()`, call `instance.save(using=using)` (the real instance) instead of `deserialized.save(using=using)` (the wrapper), and explicitly reapply `deserialized.m2m_data` afterward -- mirroring the MPTT branch immediately above it.

A fix along these lines, with regression tests, is up at #648.

Related: `netbox_custom_objects` [#689](https://github.com/netboxlabs/netbox-custom-objects/issues/689), which is where this was originally surfaced and root-caused.

bctiemann:
## Summary

#620 changed the non-MPTT delete-restore path in `ObjectChange.undo()` from `instance.save(using=using)` to `deserialized.save(using=using)`. But `DeserializedObject.save()` calls `Model.save_base(..., raw=True)` directly, which Django's own source comments say outright: *"Call save on the Model baseclass directly. This bypasses any model-defined save."* It also skips `auto_now`/`auto_now_add` population, since that's normally applied by `Field.pre_save()` during a non-raw save.

This breaks two things for any model relying on either behavior:

- A model can exclude an `auto_now` field from `serialize_object()` to keep it out of change-log diffs -- NetBox core does exactly this for `ConfigContext`'s cache fields -- and rely on `save()` to repopulate it on any real save. A raw `save_base()` leaves it as whatever the deserializer defaulted it to (typically `None` for a field missing from the serialized payload), which raises a NOT NULL violation if the column requires one.
- A model's `save()` override can perform side effects a restore still needs to run. `netbox_custom_objects.CustomObjectTypeField.save()` applies the schema DDL that (re)creates a field's physical column on creation -- a raw `save_base()` silently skips it, so reverting a field deletion restores the field's own DB row but not its actual table column.

This was originally surfaced and root-caused against [netboxlabs/netbox-custom-objects#689](https://github.com/netboxlabs/netbox-custom-objects/issues/689) (two tests failing against v1.2.0, both passing on v1.1.2 with no other changes), filed here as #650.

## Fix

Restores `instance.save(using=using)` -- the real model `.save()`, not the wrapper's raw one -- and reapplies `deserialized.m2m_data` afterward, the same pattern the MPTT branch immediately above it already uses. This keeps the M2M-preservation `#620` was after without bypassing `auto_now` population or `save()` overrides.

**Scoped down per review** to `undo()` only. Both of #689's failing tests are on `undo()`'s DELETE-restore path (reverting a deleted COT / field), and #620 -- the change that broke them -- only ever touched `undo()`. An earlier revision of this PR also modified `apply()`'s CREATE path, following a speculative finding from an automated review pass claiming the same raw-save hazard existed there. That finding doesn't hold up: `apply()`'s CREATE path was never touched by #620, isn't implicated by either of #689's failing tests, and `DeserializedObject.save()` already reapplies M2M data internally via its own `set_base()` call, so the "M2M loss" concern for `apply()` specifically doesn't actually apply. `apply()` is now byte-for-byte unchanged from `main`.

I empirically checked whether reverting `apply()` reintroduces an LTree-specific issue (the recalled original motivation for #620's raw-save switch) by running the full `netbox_branching` suite: it didn't surface one. That's not proof no such issue exists outside current test coverage, but nothing regressed.

## Testing

Three regression tests, all scoped to `undo()`'s restore path (`BaseMergeTests`, shared by both merge strategies):

- `test_merge_revert_restores_deleted_non_mptt_object` -- a non-MPTT delete+revert round trip restores the object with its M2M data (tags) intact. Note: this passes even against `main`'s original code, since `DeserializedObject.save()` already preserves M2M via its own internal mechanism -- it's general revert-with-tags coverage, not a regression guard for M2M loss specifically.
- `test_revert_restore_calls_model_save_not_raw_save_base` -- spies on `Site.save` to confirm restore actually calls the model's own `save()` method rather than bypassing it via raw `save_base()`.
- `test_revert_restore_recovers_original_created_timestamp` -- a revert-restored object's `created` recovers the true original value, sourced from the raw (un-cleaned) `prechange_data` payload rather than the already-`diff_exclude_fields()`-stripped deserialized instance.

The more production-shaped regression this fixes -- an excluded `auto_now` field causing a NOT NULL violation, and a `save()` override applying schema DDL -- doesn't have an analogous fixture within this repo's own core-model test data, so it's exercised by `netbox_custom_objects`'s own test suite instead.

Verified:
- The spy test and the timestamp test both fail against `main`'s original code and pass with this fix. The M2M test passes both before and after, for the reason noted above.
- Full `netbox_branching` suite: 408 tests, 0 new failures (1 pre-existing PostgreSQL-version-specific error in `test_upgrade_from_v4_4_10`, 1 pre-existing query-count baseline drift in `test_list_objects_with_permission` -- both confirmed to fail identically without this change).
- Full `netbox_custom_objects.tests.test_branching` suite (the one that surfaced #689): 61 tests, 0 failures.

### DIFF
--- a/netbox_branching/models/changes.py
+++ b/netbox_branching/models/changes.py
@@ -7,6 +7,7 @@ from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.postgres.fields import ArrayField
 from django.db import DEFAULT_DB_ALIAS, models
 from django.urls import reverse
+from django.utils.dateparse import parse_datetime
 from django.utils.translation import gettext_lazy as _
 from mptt.models import MPTTModel
 from netbox.models.features import ChangeLoggingMixin
@@ -36,9 +37,24 @@ def _snapshot_changelog_timestamps(instance):
     return instance.created, instance.last_updated
 
 
+def _prechange_changelog_timestamps(instance, payload):
+    """Read created/last_updated from a raw (un-cleaned) ObjectChange payload.
+    diff_exclude_fields() strips both from the *_clean view used for diffing
+    and deserializing, so a deserialized instance never has them regardless of
+    what the change actually recorded -- but the raw JSON payload still does.
+    None for a model without these fields."""
+    if not isinstance(instance, ChangeLoggingMixin):
+        return None
+    payload = payload or {}
+    return (
+        parse_datetime(payload['created']) if payload.get('created') else None,
+        parse_datetime(payload['last_updated']) if payload.get('last_updated') else None,
+    )
+
+
 def _restore_changelog_timestamps(instance, snapshot, using):
-    """Write a _snapshot_changelog_timestamps() snapshot back after a real
-    save(), via a targeted update() rather than another save()."""
+    """Write a timestamp snapshot back after a real save(), via a targeted
+    update() rather than another save()."""
     if snapshot is None:
         return
     created, last_updated = snapshot
@@ -115,29 +131,31 @@ class ObjectChange(ObjectChange_):
             # with update_fields that matches zero rows and raises NotUpdated. Applying a
             # CREATE change is always an INSERT, so force_insert is correct here. (#610)
             #
-            # A model with its own deserialize_object() hook (e.g. netbox_custom_objects')
-            # defines its own save()/M2M contract on the returned wrapper -- trust it
-            # rather than reaching into a wrapper shape it doesn't guarantee.
-            #
-            # Otherwise, every model is created through its own save(), with M2M data
-            # and change-log timestamps reapplied afterward (see the comment on the
-            # equivalent undo() path below for why: DeserializedObject.save() bypasses
-            # both, and real save() bypasses timestamps the other way).
+            # `instance.m2m_data` is Django's own DeserializedObject attribute -- its
+            # absence means the model's deserialize_object() hook returns a
+            # self-managing wrapper (e.g. netbox_custom_objects') with its own
+            # save()/M2M contract, so trust it rather than reaching into a shape it
+            # doesn't guarantee. Its presence means a plain DeserializedObject, whether
+            # from the generic deserialize_object() utility or from a hook that (like
+            # Cable's) only pre-populates extra attributes before returning one -- both
+            # need the same raw-save-bypass fix as the MPTT branch: must NOT go through
+            # DeserializedObject.save() (bypasses model-defined save() and
+            # auto_now/auto_now_add), but a real save() on an adding instance
+            # re-triggers auto_now/auto_now_add itself, so restore created/last_updated
+            # afterward too.
+            timestamps = _snapshot_changelog_timestamps(instance.object)
             if isinstance(instance.object, MPTTModel):
-                timestamps = _snapshot_changelog_timestamps(instance.object)
                 clear_mptt_fields(instance.object)
                 instance.object.save(using=using, force_insert=True)
                 for accessor_name, object_list in (instance.m2m_data or {}).items():
                     getattr(instance.object, accessor_name).set(object_list)
-                _restore_changelog_timestamps(instance.object, timestamps, using)
-            elif hasattr(model, 'deserialize_object'):
-                instance.save(using=using)
-            else:
-                timestamps = _snapshot_changelog_timestamps(instance.object)
+            elif hasattr(instance, 'm2m_data'):
                 instance.object.save(using=using)
                 for accessor_name, object_list in (instance.m2m_data or {}).items():
                     getattr(instance.object, accessor_name).set(object_list)
-                _restore_changelog_timestamps(instance.object, timestamps, using)
+            else:
+                instance.save(using=using)
+            _restore_changelog_timestamps(instance.object, timestamps, using)
 
         # Modifying an object
         elif self.action == ObjectChangeActionChoices.ACTION_UPDATE:
@@ -220,21 +238,21 @@ class ObjectChange(ObjectChange_):
             # bypassing any model-defined save() (e.g. schema DDL a save() override applies)
             # and skipping auto_now/auto_now_add. But a real save() on an adding instance
             # re-triggers auto_now/auto_now_add itself, overwriting created/last_updated
-            # with the current time instead of what the change-log payload captured (or
-            # left unset) -- hence the explicit snapshot/restore around it.
+            # with the current time -- restore the *original* values afterward, sourced
+            # from the raw prechange_data (diff_exclude_fields() strips them from
+            # prechange_data_clean, which is what `instance` was deserialized from, so
+            # they're never set on it in the first place).
+            timestamps = _prechange_changelog_timestamps(instance, self.prechange_data)
             if isinstance(instance, MPTTModel):
-                timestamps = _snapshot_changelog_timestamps(instance)
                 clear_mptt_fields(instance)
                 instance.save(using=using, force_insert=True)
                 for accessor_name, object_list in (deserialized.m2m_data or {}).items():
                     getattr(instance, accessor_name).set(object_list)
-                _restore_changelog_timestamps(instance, timestamps, using)
             else:
-                timestamps = _snapshot_changelog_timestamps(instance)
                 instance.save(using=using)
                 for accessor_name, object_list in (deserialized.m2m_data or {}).items():
                     getattr(instance, accessor_name).set(object_list)
-                _restore_changelog_timestamps(instance, timestamps, using)
+            _restore_changelog_timestamps(instance, timestamps, using)
 
     undo.alters_data = True
 
--- a/netbox_branching/models/changes.py
+++ b/netbox_branching/models/changes.py
@@ -175,20 +175,14 @@ class ObjectChange(ObjectChange_):
             # instance, which bypasses DeserializedObject's M2M handling, so — exactly as the
             # MPTT create path above does — reassign the M2M data explicitly. (#531, #610)
             #
-            # Every other model (including NetBox 4.7+ ltree models, whose path/sort_path
-            # columns are recomputed server-side by triggers) is restored the same way:
-            # through the model's own save(), with M2M data reapplied afterward. This must
-            # NOT go through DeserializedObject.save() -- that calls Model.save_base(...,
-            # raw=True) directly, which per Django's own docs "bypasses any model-defined
-            # save" and skips auto_now/auto_now_add population. Both are relied on
-            # elsewhere: a model can exclude an auto_now field from serialize_object() to
-            # keep it out of change-log diffs (NetBox core does this for ConfigContext's
-            # cache fields) and count on save() to repopulate it on any real save, and a
-            # model's save() override can perform side effects a restore still needs to
-            # run (netbox_custom_objects.CustomObjectTypeField.save() applies the schema
-            # DDL that (re)creates a field's physical column). A raw save_base() silently
-            # skips both, leaving excluded NOT NULL auto_now fields unpopulated and any
-            # such side effects un-run.
+            # Every other model is restored the same way: through the model's own save(),
+            # with M2M data reapplied afterward. Must NOT go through DeserializedObject.save()
+            # -- it calls Model.save_base(..., raw=True), which bypasses any model-defined
+            # save() and skips auto_now/auto_now_add population. Both are relied on
+            # elsewhere (e.g. an auto_now field excluded from serialize_object() to keep it
+            # out of change-log diffs, or a save() override with side effects a restore
+            # still needs), so a raw save_base() can leave excluded NOT NULL fields
+            # unpopulated or skip those side effects entirely.
             if isinstance(instance, MPTTModel):
                 clear_mptt_fields(instance)
                 instance.save(using=using, force_insert=True)
--- a/netbox_branching/models/changes.py
+++ b/netbox_branching/models/changes.py
@@ -176,17 +176,28 @@ class ObjectChange(ObjectChange_):
             # MPTT create path above does — reassign the M2M data explicitly. (#531, #610)
             #
             # Every other model (including NetBox 4.7+ ltree models, whose path/sort_path
-            # columns are recomputed server-side by triggers) is restored through
-            # DeserializedObject.save(), which persists the object and re-establishes its
-            # M2M relationships. Saving the raw instance here instead would silently drop
-            # M2M data (e.g. tags on a reverted ModuleBay).
+            # columns are recomputed server-side by triggers) is restored the same way:
+            # through the model's own save(), with M2M data reapplied afterward. This must
+            # NOT go through DeserializedObject.save() -- that calls Model.save_base(...,
+            # raw=True) directly, which per Django's own docs "bypasses any model-defined
+            # save" and skips auto_now/auto_now_add population. Both are relied on
+            # elsewhere: a model can exclude an auto_now field from serialize_object() to
+            # keep it out of change-log diffs (NetBox core does this for ConfigContext's
+            # cache fields) and count on save() to repopulate it on any real save, and a
+            # model's save() override can perform side effects a restore still needs to
+            # run (netbox_custom_objects.CustomObjectTypeField.save() applies the schema
+            # DDL that (re)creates a field's physical column). A raw save_base() silently
+            # skips both, leaving excluded NOT NULL auto_now fields unpopulated and any
+            # such side effects un-run.
             if isinstance(instance, MPTTModel):
                 clear_mptt_fields(instance)
                 instance.save(using=using, force_insert=True)
                 for accessor_name, object_list in (deserialized.m2m_data or {}).items():
                     getattr(instance, accessor_name).set(object_list)
             else:
-                deserialized.save(using=using)
+                instance.save(using=using)
+                for accessor_name, object_list in (deserialized.m2m_data or {}).items():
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review and limit response to actionable issues.

bctiemann:
Thanks for the review.

**Finding 1 (apply() CREATE path)**: fixed, but it turned out to need more care than "cheap and consistent" suggested. A first pass calling `instance.object.save()` unconditionally broke `netbox_custom_objects`'s own models: their `deserialize_object()` classmethod hook returns a custom wrapper (`_SchemaAwareDeserialized`) that was *already* working around this exact hazard on its own -- its `save()` runs the real model `save()` and reapplies M2M data internally, and it doesn't expose a `.m2m_data` attribute the way a plain `DeserializedObject` does. Calling `instance.object.save()` directly skipped that wrapper's own logic entirely, and the subsequent `instance.m2m_data` access raised `AttributeError` -- caught by running `netbox_custom_objects`'s branching suite, which errored on exactly that. Fixed version now branches on `hasattr(model, 'deserialize_object')`: the hook branch keeps calling `instance.save(using=using)` unchanged (trusting whatever contract the model itself defines), and only the plain-utility branch gets the `instance.object.save()` + M2M-reapply treatment. Added a mirroring regression test (`test_merge_apply_create_calls_model_save_not_raw_save_base`) and re-verified against both suites.

**Finding 2 (`mock_save.called` vs `call_count >= 1`)**: these are equivalent (`Mock.called` is `True` iff `call_count >= 1`), so no behavior to change there -- leaving as `.called` since it reads slightly more directly for "was this invoked at all."

PR description and verification numbers updated.

bctiemann:
Note the last commit adds significant docstring volume to explain the `elif` branch and classmethod hook, but I think it's justified in this case.

arthanson (changes requested):
This looks like it will reset the created and last_updated when you merge and revert - it looks like this was partially broken before. Previously it should keep the created when you did the merge or revert.

What It should be:
merge a branch-created Site:  created = 15:59:51.612000   last_updated = None
revert a branch delete:       created = 15:59:51.612000   last_updated = None

PR #648 
merge a branch-created Site:  created = merge time      last_updated = merge time
revert a branch delete:       created = revert time     last_updated = revert time

Also Claude brought up:

undo() never honors the deserialize_object hook - apply() branches on it (changes.py:76) but undo()'s restore always uses the plain utility (changes.py:171), so revert still bypasses netbox_custom_objects' schema-aware wrapper - the same skipped-save() bug this PR fixes.

bctiemann:
Thanks for catching this -- confirmed, both findings.

**created/last_updated reset on merge/revert**: verified empirically by printing the actual `postchange_data`/`prechange_data` JSON payloads directly (not just inferring from behavior). `created` is genuinely present in `postchange_data` and should be preserved on merge; `last_updated` is genuinely absent (`CHANGELOG_SKIP_EMPTY_CHANGES` excludes it from serialization) and should end up `None`, matching pre-#647 (raw `save_base()`) behavior exactly. The regression: a real, non-raw `save()` on an adding instance re-triggers Django's `auto_now`/`auto_now_add` handling, which overwrites both fields with the current apply/undo wall-clock time regardless of what the payload provided.

Fixed by snapshotting `created`/`last_updated` immediately before the real `save()` call and writing them back with a targeted column update immediately after, in all four affected branches (both MPTT and non-MPTT, in both `apply()`'s CREATE and `undo()`'s DELETE-restore). Verified with two new tests (on both merge strategies) that fail against the pre-fix code and pass with it.

One caveat surfaced along the way, worth flagging explicitly: on revert, `created` still ends up `None`, not the object's *original* creation time. That's a separate, pre-existing gap (confirmed present in v1.1.2 too, unrelated to raw-vs-real save): `undo()`'s restore path deserializes from `prechange_data_clean`, and `ObjectChange.diff_exclude_fields()` (NetBox core) strips `created`/`last_updated` from that specifically for diff-display purposes -- so the original timestamp isn't available to restore in the first place. This PR restores parity with pre-#647 behavior (`None`, not a wrong revert-time value) but doesn't attempt to close that deeper gap, which would mean deserializing from the un-cleaned `prechange_data` instead -- a bigger, riskier change I didn't want to fold in here without a closer look at what else that data source affects (MPTT/ltree bookkeeping columns are also stripped by the same mechanism).

**`undo()` never checks the `deserialize_object` hook**: looked into wiring this in to match `apply()`, but backed off after checking `CustomObjectType.deserialize_object()`'s own docstring in netbox_custom_objects -- it explicitly scopes itself to "`apply()` hook for CREATE actions" (including CREATE-specific squash-ordering logic for field-before-COT replay ordering), and its wrapper's `save()` has no timestamp-preservation logic of its own (it would need the same fix I just made, but on that repo's side). A quick empirical test confirmed forcing the hook into `undo()` doesn't obviously break anything today, but I don't think it's my call to make from this side -- extending that hook's contract to also cover restore semantics belongs with `netbox_custom_objects`'s own maintainers. In practice this isn't costing correctness today: `undo()`'s existing plain-utility path already dispatches to a model's own `save()` override correctly, since `instance = deserialized.object` is always the real, unwrapped instance regardless of which deserializer built it -- confirmed by the full `netbox_custom_objects.tests.test_branching` suite (61 tests, including exactly this scenario) passing cleanly.

PR description updated with full details and verification numbers.

netbox_branching/models/changes.py
```diff
@@ -92,13 +114,30 @@ def apply(self, branch, using=DEFAULT_DB_ALIAS, logger=None, skip_missing=False)
...
             if isinstance(instance.object, MPTTModel):
+                timestamps = _snapshot_changelog_timestamps(instance.object)
                 clear_mptt_fields(instance.object)
                 instance.object.save(using=using, force_insert=True)
                 for accessor_name, object_list in (instance.m2m_data or {}).items():
                     getattr(instance.object, accessor_name).set(object_list)
-            else:
+                _restore_changelog_timestamps(instance.object, timestamps, using)
+            elif hasattr(model, 'deserialize_object'):
                 instance.save(using=using)
```
arthanson: probably need the timestamps setting here and if so can pull that up to the top before the first if statement (also put the restore at the bottom after all the if cases?)

netbox_branching/models/changes.py
```diff
@@ -175,18 +214,27 @@ def undo(self, branch, using=DEFAULT_DB_ALIAS, logger=None):
...
             if isinstance(instance, MPTTModel):
+                timestamps = _snapshot_changelog_timestamps(instance)
                 clear_mptt_fields(instance)
                 instance.save(using=using, force_insert=True)
                 for accessor_name, object_list in (deserialized.m2m_data or {}).items():
                     getattr(instance, accessor_name).set(object_list)
+                _restore_changelog_timestamps(instance, timestamps, using)
             else:
-                deserialized.save(using=using)
+                timestamps = _snapshot_changelog_timestamps(instance)
```
arthanson: again can hoist the timestamp stuff outside the if statement

netbox_branching/models/changes.py
```diff
@@ -92,13 +114,30 @@ def apply(self, branch, using=DEFAULT_DB_ALIAS, logger=None, skip_missing=False)
...
+            # both, and real save() bypasses timestamps the other way).
             if isinstance(instance.object, MPTTModel):
+                timestamps = _snapshot_changelog_timestamps(instance.object)
                 clear_mptt_fields(instance.object)
                 instance.object.save(using=using, force_insert=True)
                 for accessor_name, object_list in (instance.m2m_data or {}).items():
                     getattr(instance.object, accessor_name).set(object_list)
-            else:
+                _restore_changelog_timestamps(instance.object, timestamps, using)
+            elif hasattr(model, 'deserialize_object'):
```
arthanson: The comment above 'A model with its own deserialize_object() hook (e.g. netbox_custom_objects')' I think is misleading as that is doing the old default action.  Also I think this has to be looked at more closely as Cable model in NetBox as a deserialize_object - so I think this is catching more then intended?  Different more specific check might be needed here?

Also doing instance.object.save will cause things like VirtualMachine.save() now runs apply_type_defaults() on the deserialized instance; ltree models re-enter the trigger-refresh path - model save() overrides fire during merge for the first time ever, for every non-MPTT model without a deserialize_object hook. Note: the Cable is an exception as it is caught in the 'deserialize_object' check. Not sure that is intended and potential of a lot of side-effects.

arthanson (changes requested):
For undo case - The original timestamp is in the change record - diff_exclude_fields() only strips it from the _clean view used for diff display. undo() doesn't need to change what it deserializes from; it only needs to source the two timestamp values from the un-cleaned payload instead of from the deserialized instance:

  timestamps = (
      (self.prechange_data or {}).get('created'),
      (self.prechange_data or {}).get('last_updated'),
  )

bctiemann:
Both fixed, thanks -- and confirmed the hoisting too.

**Cable / `deserialize_object` catching more than intended**: confirmed. `Cable.deserialize_object()` returns a plain `DeserializedObject` (it exists only to pop `a_terminations`/`b_terminations` before deserializing), not a self-managing wrapper -- so keying the dispatch off `hasattr(model, 'deserialize_object')` was wrong; it routed Cable into the "trust it, don't fix it" branch by accident. Switched to `hasattr(instance, 'm2m_data')` -- Django's own `DeserializedObject` attribute, present on Cable's result and the plain utility's, but deliberately absent from `netbox_custom_objects`'s wrapper. That's a structural, instance-level check rather than a name-based, model-level one, so a future model defining `deserialize_object()` for its own unrelated reasons can't get miscategorized the same way again.

I dug into whether this was actually causing an observable bug for Cable and it wasn't: I created a Cable in a branch, merged, and checked `CableTermination`/`Interface.cable_id` linkage both before and after this fix -- identical, correct results either way. Turns out those side effects are independently change-tracked via their own `ObjectChange` records from the original branch-side `Cable.save()`, so they get correctly replayed regardless of whether `Cable.save()`'s override re-fires during the Cable object's *own* replay. Added a regression test for it anyway (`test_merge_cable_created_in_branch_has_terminations`), since the dispatch fix is correct on its own terms even though it's not fixing an observed bug for this specific model.

On your broader flag -- **agreed, and I want to be upfront that I can't fully close this out myself.** Calling a model's real `save()` during `apply()`'s CREATE path (for any model without a self-managing wrapper) genuinely is new behavior: this path has *always* used raw `save_base()` for such models, even before #647 (which never touched `apply()`). I checked `VirtualMachine.apply_type_defaults()` specifically -- it only fills fields that are still `None`, so it's a no-op for a fully-populated deserialized instance in the common case -- but I haven't audited every model in core or every plugin for a `save()` override with a side effect that *isn't* independently changelogged the way Cable's is. The full `netbox_branching` suite (which exercises a lot of core models through merge/apply/revert scenarios) passes cleanly, which is reassuring but not exhaustive. I don't think it's my call to decide "this is fine, ship it" for a change with this blast radius -- want me to file a tracking issue for a broader audit, or would you rather look at scoping this PR's fix down (e.g. an explicit opt-in marker instead of blanket-flipping every model), or something else? Happy to do either, just don't want to unilaterally decide the risk is acceptable.

**`undo()`'s timestamp fix wasn't actually recovering anything**: confirmed and fixed exactly as you suggested. My snapshot was reading from the deserialized `instance`, which is built from `prechange_data_clean` -- already stripped of `created`/`last_updated` by `diff_exclude_fields()` -- so it was always capturing `(None, None)`. That happened to match old pre-#647 behavior, but by coincidence, not because it was recovering anything. Now sources both fields from the un-cleaned `self.prechange_data` payload instead (parsed via `parse_datetime`, since the raw JSON has them as ISO strings), which does still have them. Updated the existing test (`test_revert_restore_recovers_original_created_timestamp`) to assert the real original value comes back, rather than asserting `None`.

Updated PR description with full details. All tests pass (414 branching, 61 netbox_custom_objects).

bctiemann:
@claude /review and focus on changes since the original review, considering validity of requested changes and arrived-at solutions. Look at adjacent behavior changes if any.

bctiemann:
Scoped this down to just what is needed to fix the `undo()` breakage from #620.

Backed-out all `apply()` changes.

On the question of LTree -- ran the full suite with `apply()` fully reverted and `undo()` back to `instance.save()`: **nothing broke.** 408 tests, same 2 pre-existing unrelated failures as before, no new LTree-specific failures. That's obviously not proof there's no LTree issue outside current test coverage, but empirically nothing in the current suite surfaces one from this change. If you remember more specifically what broke originally (a particular model, a particular sequence), I'm happy to try to reproduce it directly -- otherwise this seems safe to proceed with as-is, and we can revisit if something concrete turns up.

Removed the `apply()`-related tests.
Kept the three `undo()`-scoped tests; re-verified all three against `main`'s original code -- two fail as expected, and one (the M2M-preservation test) passes even without the fix, which I've noted in both the test and PR description: `DeserializedObject.save()` already preserves M2M via its own mechanism, so that test was always general revert-with-tags coverage, not a regression guard for M2M loss specifically.

PR description rewritten (not just appended) to reflect the final, scoped-down state.

netbox_branching/models/changes.py
```diff
@@ -175,18 +205,27 @@ def undo(self, branch, using=DEFAULT_DB_ALIAS, logger=None):
...
+            # and skipping auto_now/auto_now_add. But a real save() on an adding instance
+            # re-triggers auto_now/auto_now_add itself, overwriting created/last_updated
+            # with the current time -- restore the *original* values afterward, sourced
+            # from the raw prechange_data (diff_exclude_fields() strips them from
+            # prechange_data_clean, which is what `instance` was deserialized from, so
+            # they're never set on it in the first place).
+            timestamps = _prechange_changelog_timestamps(instance, self.prechange_data)
             if isinstance(instance, MPTTModel):
                 clear_mptt_fields(instance)
                 instance.save(using=using, force_insert=True)
```
arthanson, line 221: The whole block instance.save and for accessor_name can be hoisted below so the if statement only does the clear_mptt_fields.

arthanson (changes requested):

arthanson (approved):

PR #652, opened 2026-09-03
611 - add regression tests for front/rear port mappings merge/sync

### LINKED ISSUE #611: Front/rear port mappings are silently dropped on branch merge (NetBox 4.5+)
### Plugin Version

1.0.4

### NetBox Version

4.5.10

### Python Version

3.12

### Steps to Reproduce

1. On NetBox 4.5.x with netbox-branching installed, create and activate a branch.
2. In the branch, create a device.
3. Add a rear port to the device.
4. Add a front port on the device mapped to that rear port (front-port position → rear-port position).
5. Confirm in the branch that the front port shows its rear-port mapping (rear_ports populated).
6. Merge the branch with commit.

### Expected Behavior

After merge, main contains the device, the rear port, the front port, and the front↔rear mapping. The front port's rear_ports (and the rear port's front_ports) are populated in main, matching what was in the branch.

### Observed Behavior

The device, rear port, and front port all merge into main, but the front↔rear mapping is silently dropped. In main the front port's rear_ports is [] and the rear port's front_ports is []. The merge reports success with no error and no conflict. The ports arrive unmapped.

arthanson:
### Fixes: #611 

The issue for #611 was actually fixed in NetBox, but this adds regression tests in branching to catch any regressions in the future.

### DIFF
--- a/netbox_branching/tests/test_iterative_merge.py
+++ b/netbox_branching/tests/test_iterative_merge.py
@@ -1326,12 +1326,16 @@ class BaseMergeTests:
 
         # In branch: map the existing front port to the existing rear port
         with activate_branch(branch), event_tracking(request):
-            PortMapping.objects.create(
+            mapping = PortMapping.objects.create(
                 front_port=FrontPort.objects.get(id=front_port_id),
                 rear_port=RearPort.objects.get(id=rear_port_id),
                 front_port_position=3,
                 rear_port_position=4,
             )
+            mapping_id = mapping.id
+
+        # The mapping must be change-logged for the merge to be able to replay it
+        self._assert_object_changes(branch, PortMapping, mapping_id, 1, ['create'])
 
         # Main is still unmapped prior to the merge
         self.assertEqual(PortMapping.objects.filter(front_port_id=front_port_id).count(), 0)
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
Per Claude review:

invalid field - this is incorrect, no change needed

ObjectChange assertion - change made.

netbox_branching/tests/test_iterative_merge.py
```diff
@@ -1236,6 +1239,158 @@ def test_merge_cable_path_recalculation(self):
...
+                rear_port=rear_port,
+                front_port_position=1,
+                rear_port_position=2,
+            )
+            front_port_id = front_port.id
+            rear_port_id = rear_port.id
+            mapping_id = mapping.id
+
+        # The mapping must be change-logged for the merge to be able to replay it
+        self._assert_object_changes(branch, PortMapping, mapping_id, 1, ['create'])
```
jeremystretch, line 1290: Should we add an assertion to check that the port mapping isn't in main prior to the merge (as in the other two tests)?
```suggestion
        self.assertEqual(PortMapping.objects.filter(front_port_id=front_port_id).count(), 0)
        self._assert_object_changes(branch, PortMapping, mapping_id, 1, ['create'])
```

netbox_branching/tests/test_iterative_merge.py
```diff
@@ -1236,6 +1239,158 @@ def test_merge_cable_path_recalculation(self):
         # A successful cable connection creates two CablePath records (one per endpoint)
         self.assertEqual(CablePath.objects.count(), 2, 'Cable paths not populated after merge')
 
+    def _create_ports(self, device, suffix=''):
```
jeremystretch: `suffix` appears to be unused; remove it?

jeremystretch (changes requested):

jeremystretch (approved):

PR #656, opened 2026-09-10
Fix #632 - Distinguish merge collisions with main from invalid branch values

### LINKED ISSUE #632: Branch merge fails with position ValidationError when branch and main independently occupy the same rack slot
### Plugin Version

1.1.2

### NetBox Version

4.5.10

### Python Version

3.12.3

### Steps to Reproduce

1. In main, create a site, a rack, and a 1U device type with a device role.
2. Create a branch.
3. In the branch, create device A in the rack at position 12, face front.
4. In main, create device B in the same rack at position 12, face front. Both creates are valid in their own schema, so neither is rejected.
5. Merge the branch (iterative strategy).

### Expected Behavior

The rack unit collision is surfaced as a conflict before the merge starts, or the merge fails with a report that gives the user a way to resolve it from within the branch.

### Observed Behavior

No conflict is detected before the merge. Conflict detection compares field values on the same object, and here the collision is between two independent objects, so there is nothing to flag. The merge job starts and errors while replaying the branch CREATE:

ValidationError({'position': ['U12.0 is already occupied or does not have sufficient space to accommodate this device type: example-dt (1.0U)']})

The job report contains a single validation_error entry on field position for the branch device. Retrying the merge fails identically. The branch returns to Ready status but can never merge: the colliding object lives in main, so no edit inside the branch resolves it short of deleting the branch device and losing the work. The practical workaround is to redo the change in a new branch after checking main, which defeats long-lived branches.

This is related to but distinct from netboxlabs/netbox-branching#624. That issue covers rack/position/face divergence on a single object. Here two different objects collide on rack unit uniqueness, and PR netboxlabs/netbox-branching#625's conflict gating does not cover it because there is no per-object divergence to detect.

arthanson:
### Fixes: #632 

This is a valid conflict that the user must fix, so made a specific check for this in the merge report. photo below.  For merge even if you fix you must use squash merge as iterative merge will still have an try to replay the original conflicting ObjectChange.

When a branch object and a main object independently claim the same resource (e.g. the same rack unit), no conflict is detectable up front and the merge reported it as an invalid value in the branch - misleading, since the branch value is valid and the colliding object lives in main.

Merge validation failures are now re-checked against the branch's own schema. If the failure doesn't reproduce there, it's reported as a collision with main, quoting the underlying error and offering two real routes: resolve it in main, or change the value in the branch and merge with squash (iterative replays the original colliding value, so it fails again).

Reporting only - merge behavior is unchanged, and sync is untouched.

Monosnap Issue 632B Rack Collision - Merge Report | NetBox 2026-09-10 14-29-28

### DIFF
--- a/docs/using-branches/syncing-merging.md
+++ b/docs/using-branches/syncing-merging.md
@@ -122,16 +122,14 @@ Alternatively, if the conflicting changes are problematic, you can go back and m
 
 ### Collisions With Objects in Main
 
-Conflicts are detected by comparing the *same* object in your branch and in main, so a collision between two *different* objects cannot be flagged in advance. The common example is a shared resource: you place a device in rack unit 12 inside your branch, and meanwhile someone places a different device in that same slot in main. Each write is perfectly valid in its own schema, nothing is flagged as a conflict, and the collision only surfaces when the merge replays your change against main.
+Conflicts are detected by comparing the same object in your branch and in main, so a collision between two *different* objects cannot be flagged in advance — for example, a device you place in rack unit 12 in your branch and a different device someone places in that slot in main. Both writes are valid in their own schema; the collision surfaces only when the merge replays your change against main.
 
-When this happens, the merge fails and the job report describes it as a collision with the main schema, naming the underlying validation error (for example, "U12 is already occupied…"). This is worth distinguishing from an ordinary validation error, because the object you are colliding with is not in your branch at all — it is not visible there and no edit you make inside the branch will make it go away. There are two ways forward:
+The job report identifies these as a collision with the main schema and quotes the underlying error ("U12 is already occupied…"). Two ways forward:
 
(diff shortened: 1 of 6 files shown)


arthanson:

arthanson:
Per Claude review:

 1. Fixed - probe now requires the branch object to validate cleanly; field-scoped check was unsound since `clean()` short-circuits.

2. Fixed - `_classify_validation_error` maps `NON_FIELD_ERRORS` to `None`, so the sentinel never reaches messages, recommendations, or the Field column.

jeremystretch (changes requested):
I'm concerned about the volume of logic introduced here just to improve an error message. The initial review also returned a few concerns that would require additional work.

I've submitted a second PR (#663) which proposes significant simplification of the approach taken here. Please have a look and see if I've missed anything.

jeremystretch (approved):

PR #657, opened 2026-09-10
Convert to use Plugin Scaffolding

arthanson:
Adopts [netbox-plugin-scaffold](https://github.com/netboxlabs/netbox-plugin-scaffold) as this repo's baseline, tracked via `.copier-answers.yml` for future `copier update`s.

- **Tooling** - `ruff.toml` folded into `pyproject.toml`; pre-commit expanded to the full scaffold stack; added `.yamllint`, `.github/labels.yml`.
- **CI** -`lint-tests.yaml` → `test.yml` (pre-commit gate, coverage leg, v4.7.0 floor); `claude.yaml` → `claude-review.yml`.
- **Formatting** - repo-wide `ruff format` + djLint. See `.git-blame-ignore-revs`.
- **Not adopted** - the scaffold's CodeArtifact release flow. `release.yaml`, `scripts/`, the PyPI distribution name and `testing/configuration.py` are unchanged. Carve-outs listed in `AGENTS.md`.

Full suite passes against v4.7.0 with no regressions; wheel still passes the release checks.

3 CodeQL alerts on `test.yml`, caused by tightening `permissions` to `contents: read` (the old workflow's `contents: write` put it outside the rule's scope). The scaffold's own `test.yml` flags identically. Needs a call on fix vs. dismiss.


arthanson:

arthanson:
Per Claude review:

1. Fixed - dropped the redundant ||; it bypassed the regex guard. Won't affect CodeQL: ddef02f used this exact form and flagged.
2. Not fixed - scaffold-owned, unused is intentional. Will need to look at possibly adopting it in a follow on PR.
3. Not valid - cache removed in 03e1bc1 and 7087565, alerts unchanged. Cause is permissions: contents: read.

PR #659, opened 2026-09-14
Test Performance Improvements

arthanson:
Two changes targeting fixed per-test overhead, which was ~93% of branch-test execution time.

**Faster `TransactionTestCase` teardown** - Django TRUNCATEs all 211 tables after every test, however few rows were written. We now find the non-empty ones in a single query and DELETE just those. Sequences are unaffected; falls back to Django's flush if not permitted.

**Drop leaked branch schemas at teardown** - the flush deletes Branch rows directly, bypassing the deprovision that delete() would do, leaking ~150 tables and ~1,000 indexes per test into pg_class and slowing every later provision (−35% on a 172-test chunk).

### Measured (local, clean DB)

| | before | after |
|---|---|---|
| teardown | 1.69s | ~0.6s (flush 0.02s + schema drop) |
| suite | 160.1s | 77.7s (**−51%**) |

Expect ~27% off the CI test phase. The ~263s test-DB migration step is unaffected.


arthanson:

arthanson:
Per Claude review:

**_make_batch_task / autocommit:** Fixed - dropped BEGIN/COMMIT; Postgres already wraps multi-statement queries in an implicit transaction, so it's correct in either mode.

**_fixture_teardown fallback:** Fixed - now falls back per alias via _django_flush(db_name), so already-flushed aliases aren't TRUNCATEd twice.

arthanson:

arthanson:
Per Claude:

1. fixed
2. fixed
3. acceptable minor version jumps are okay.

jnovinger:
@arthanson , apologies, this now has merge conflicts (I think because of changes I just merged in #662).

jnovinger (approved):

PR #660, opened 2026-09-15
#640 - Advance the conflict baseline when syncing a branch

### LINKED ISSUE #640: Sync does not advance the conflict baseline
### Plugin Version

1.1.3

### NetBox Version

4.6.9

### Python Version

3.14.4

### Steps to Reproduce

A device is being taken out of service while the data centre team is working on it in main.

1. In main, take a device which is assigned to a rack and whose status is **Active**. The example below uses `dmi01-akron-rtr01`, rack 1.
2. Create a branch and open it.
3. In the branch, clear the device's rack, to prepare it for stock. Leave the status alone.
4. In main, outside the branch, set the same device's status to **Offline**.
5. Sync the branch. The device in the branch now reads status **Offline**, rack empty.
6. In the branch, set the device's status to **Inventory**.
7. Open the branch diff for that device, then try to merge the branch.

Every branch edit to `status` happens at step 6, after the sync at step 5 has taken main's value. Main changes nothing after step 4.

### Expected Behavior

No conflict on `status`, and the branch merges.

The branch synced main's change before making its own. In git this is a rebase followed by one commit, and it fast-forwards.

### Observed Behavior

`status` is reported as a conflict. The resulting `ChangeDiff`:

| Snapshot | `status` | `rack` |
|---|---|---|
| `original`, the baseline | `active` | `1` |
| `modified`, the branch | `inventory` | `None` |
| `current`, main now | `offline` | `1` |

`conflicts == ['status']`.

The merge is then refused:

```
All conflicts must be acknowledged in order to merge the branch.
```

Syncing again does not clear it. `Branch.sync()` returns early when `get_unsynced_changes()` is empty, and even a sync that does run never touches `original`. The flag stands for the life of the branch, and the only way past it is to tick the acknowledgement box by hand.

`rack` shows the check is not simply broken: main never touched it, so the branch clearing it raises nothing.

### Suspected Cause

The underlying question: should a sync behave like a rebase, and move the baseline the diff is computed against?

`ChangeDiff` holds three snapshots: `original`, `modified` (the branch's state) and `current` (main's state now). `_update_conflicts()` runs a three-way test on every field of an updated object:

```python
if self.action == ObjectChangeActionChoices.ACTION_UPDATE:
    ...
    conflicts = [
        k for k, v in original.items()
        if v != modified.get(k) and v != current.get(k) and modified.get(k) != current.get(k)
    ]
```

The comparison is correct. A field is flagged only when the branch moved it away from the baseline, main moved it away too, and the two landed on different values. The fault is in the baseline it reads.

`original` is written exactly once, when the `ChangeDiff` is created, from `prechange_data_clean` of the **branch's first change to that object**. Nothing else ever assigns it. A sync updates `modified`, through the synthetic `ObjectChange` written by `_flush_sync_buffer()`, but never the baseline. Main's change therefore keeps reading as divergence, even though the branch has absorbed it.

Swap steps 3 and 5 and the problem disappears: the `ChangeDiff` is then created from a prechange state which already holds main's value, so `original == current` and nothing is flagged. The baseline is therefore already meant to be main's state. It is simply not maintained when the branch touches the object first.

The acknowledgement warning shows the same assumption:

> The following N conflicts were found. Applying these changes will overwrite data that has been modified since the branch was created. Please confirm that each of these changes is desired.

The sentence is true. Main did change `status` after the branch was created. It is still not a conflict: the branch synced that change, saw `offline`, and chose `inventory` on top of it. Once a branch has synced, the window that matters is *since the last sync*, and in that window main changed nothing.

Had the sync moved `original` to main's state at that point, `status=offline, rack=1`, the same three-way test returns the right answer with no other change:

| Field | `original` | `modified` | `current` | Result |
|---|---|---|---|---|
| `status` | `offline` | `inventory` | `offline` | `original == current`, no conflict |
| `rack` | `1` | `None` | `1` | `original == current`, no conflict |

Moving the baseline does not hide a real conflict. Add a step 8 in which main assigns the device to rack 2, after the branch has synced. The branch holds no rack, main holds rack 2, and the new baseline holds rack 1. All three differ, so `rack` is flagged, which is correct: two people changed the same field and neither saw the other's value.

The baseline must become **main's state at sync time**, not the branch's state after the sync. Using the branch's post-sync state would fold the branch's own edits into the baseline, and a later change in main to a field the branch had already changed would then go unnoticed.

### Proposed Fix

When a sync applies a change from main to an object, move that object's baseline with it. Set `original` to the state main had at that point, then recompute the conflicts.

- **Which value.** The `postchange_data_clean` of the last change from main replayed for that object. `_apply_sync_update()` already has it. The `sync_buffer` does not: it holds the branch's states, not main's.
- **Where to do it.** In the same transaction as `_flush_sync_buffer()`, so a dry run undoes the new baseline as well.
- **What to leave alone.** Objects whose change from main could not be applied, for example one the branch has deleted. Moving their baseline would hide a real conflict.

arthanson:
### Fixes: #640 

Syncing takes main's changes, but the "original" snapshot conflict detection compares against stayed frozen at the branch's first edit. So syncing a field and then editing it yourself was flagged as a conflict forever, clearable only by hand.

Sync now behaves like a rebase: main's state is written into the baseline once the sync commits. Real conflicts still surface, branch-deleted objects are skipped, and dry runs change nothing.

Adds six regression tests and clarifies the docs on "Original".


arthanson:

arthanson:
Per Claude review, 1 and 2 fixed.

PR #661, opened 2026-09-15
#312 - Enforce object-based permissions on branches

### LINKED ISSUE #312: Branch Permissions
### Plugin Version

0.6.1

### Proposed functionality

User rights should be adjustable so that the drop-down menu only displays the branches to which the user has access. And in the side menu, the “Branching” menu item should be hidden if the user does not have rights for the plugin.

The attached image shows a test user who only has the right to modify some IPAM objects. However, he can still see the "Branching" menu item and can also view and select existing branches and make changes to the IPAM objects in the braches.

Image

### Use case

I don't want users to be able to make changes in foreign branches. Ideally, branches that they don't have access to or haven't created themselves should not appear in the selection list. This is to prevent changes from being made or objects from being deleted in my test branch, for example.

### External dependencies

_No response_

arthanson:
### Fixes: #312 

Filters branches through NetBox's object-based permissions everywhere they're exposed: the navbar selector, branch activation, the REST API, and bulk migrate. Admins scope branches with ObjectPermission constraints (e.g. `{"owner": "$user"}`).

Branch resolution runs in middleware, before DRF authenticates, so token-authenticated callers are identified via `resolve_request_user()`.

Adds `docs/permissions.md` and `tests/test_permissions.py`. Existing branch users now need `view_branch` granted explicitly.

Branch menu from Admin:
Monosnap Home | NetBox 2026-09-15 14-19-47

Same branch menu as user Alice without permissions for branch Bob
Monosnap Home | NetBox 2026-09-15 14-18-21


arthanson:

arthanson:
Per Claude:
1. Fixed - _UNRESOLVED sentinel replaces the is not None cache guard.
2. Partly fixed - enumeration closed via restrict() (404 for both cases); view-guard deliberately not added, since NetBox scopes each action independently.
3. Documented, not fixed - table row now says migrate is UI-only; adding the endpoint is a separate feature.
4. Not warranted - alerts() runs once per detail page, and the queryset is a lazy subquery over an already-cached perm lookup.

netbox_branching/utilities.py
```diff
@@ -535,22 +543,73 @@ def is_api_request(request):
...
+    eligible for activation.
     """
+    # Resolved lazily: a request which names no branch should not incur the cost of identifying the user.
+    branches = SimpleLazyObject(lambda: get_branches_for_user(resolve_request_user(request)))
+
     # The active Branch may be specified by HTTP header for REST & GraphQL API requests.
-    from .models import Branch
     if is_api_request(request) and BRANCH_HEADER in request.headers:
-        branch = Branch.objects.get(schema_id=request.headers.get(BRANCH_HEADER))
+        branch = branches.get(schema_id=request.headers.get(BRANCH_HEADER))
```
jeremystretch: Can we just call `get_branches_for_user()` with `request.user` directly here and avoid complicating it with a lazy object?
```suggestion
        branch = get_branches_for_user(request.user).get(schema_id=request.headers.get(BRANCH_HEADER))
```

netbox_branching/utilities.py
```diff
@@ -535,22 +543,73 @@ def is_api_request(request):
...
         if not branch.ready:
             return HttpResponseBadRequest(f"Branch {branch} is not ready for use (status: {branch.status})")
         return branch
 
     # Branch activated/deactivated by URL query parameter
     if QUERY_PARAM in request.GET:
         if schema_id := request.GET.get(QUERY_PARAM):
-            branch = Branch.objects.get(schema_id=schema_id)
+            branch = branches.get(schema_id=schema_id)
```
jeremystretch:
```suggestion
            branch = get_branches_for_user(request.user).get(schema_id=schema_id)
```

netbox_branching/utilities.py
```diff
@@ -574,7 +633,7 @@ def get_active_branch(request):
     # Branch set by cookie
     if schema_id := request.COOKIES.get(COOKIE_NAME):
         try:
-            branch = Branch.objects.get(schema_id=schema_id)
+            branch = branches.get(schema_id=schema_id)
```
jeremystretch:
```suggestion
            branch = get_branches_for_user(request.user).get(schema_id=schema_id)
```

netbox_branching/utilities.py
```diff
@@ -535,22 +543,73 @@ def is_api_request(request):
...
+
 def get_active_branch(request):
     """
-    Return the active Branch (if any).
+    Return the active Branch (if any). Only branches which the requesting user is permitted to view are
+    eligible for activation.
     """
+    # Resolved lazily: a request which names no branch should not incur the cost of identifying the user.
+    branches = SimpleLazyObject(lambda: get_branches_for_user(resolve_request_user(request)))
+
```
jeremystretch:
```suggestion
```

netbox_branching/utilities.py
```diff
@@ -535,22 +543,73 @@ def is_api_request(request):
...
+            authenticators=[auth() for auth in api_settings.DEFAULT_AUTHENTICATION_CLASSES],
+        ).user
+    except Exception:
+        # Leave the user unresolved; the view will report the authentication failure itself
+        logger.debug('Unable to authenticate API request while resolving the active branch', exc_info=True)
+    request._branching_api_user = user
+
+    return user
+
+
```
jeremystretch, line 588: Not needed if we simplify the call to `restrict()` on the queryset. (Also, recreating the user instance from a new `DRFRequest` is breaking per a Claude review.)
```suggestion
```

netbox_branching/api/views.py
```diff
@@ -3,15 +3,22 @@
 from core.api.serializers import JobSerializer
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseBadRequest
+from django.shortcuts import get_object_or_404
```
jeremystretch: Use DRF's `get_object_or_404()` for API requests. (Also need to re-order the imports).
```suggestion
from rest_framework.generics import get_object_or_404
```

netbox_branching/utilities.py
```diff
@@ -535,22 +543,73 @@ def is_api_request(request):
...
+    from rest_framework.request import Request as DRFRequest
+    from rest_framework.settings import api_settings
+
+    try:
+        user = DRFRequest(
+            request,
+            authenticators=[auth() for auth in api_settings.DEFAULT_AUTHENTICATION_CLASSES],
+        ).user
+    except Exception:
```
jeremystretch: Only if we need to keep `resolve_request_user()`:

With an expired, disabled, or revoked token plus an `X-NetBox-Branch` header: `AuthenticationFailed` is swallowed and logged at DEBUG, `user` stays `AnonymousUser`, `Branch.objects.restrict(anon, 'view')` returns `.none()`, `branches.get(...)` raises `Branch.DoesNotExist`, and `BranchMiddleware.__call__` returns `HttpResponseBadRequest("Invalid branch identifier")`. The caller gets a 400 saying the *branch id* is wrong and never reaches DRF, so they never see the 401 explaining the real problem. The docstring's claim that "the view will report the authentication failure itself" doesn't hold once a branch is named.

Fix: catch `rest_framework.exceptions.APIException` specifically and let the branch lookup fall through to `None` so DRF can report the 401.

docs/permissions.md
```diff
@@ -0,0 +1,71 @@
...
+| Action | Permission | Description |
+|---|---|---|
+| `sync` | `netbox_branching.sync_branch` | Pull changes from main into the branch |
+| `merge` | `netbox_branching.merge_branch` | Apply the branch's changes to main |
+| `revert` | `netbox_branching.revert_branch` | Undo a merged branch's changes |
+| `migrate` | `netbox_branching.migrate_branch` | Apply outstanding migrations to the branch schema (UI only — the REST API exposes no migrate endpoint) |
+| `archive` | `netbox_branching.archive_branch` | Deprovision a merged branch's schema |
+
+A user who has not been granted `view` on any branch sees neither the **Branching** navigation menu nor the branch
+selector in the header, and cannot activate a branch.
```
jeremystretch: Technically, a user with permission to view ChangeDiffs will still see the navigation menu.

docs/permissions.md
jeremystretch, line 1: It would be cleaner to split this into a separate documentation PR, as it exceeds the scope of a bug report.

netbox_branching/template_content.py
```diff
@@ -17,9 +18,15 @@
 class BranchSelector(PluginTemplateExtension):
 
     def navbar(self):
+        user = self.context['request'].user
+
+        # Hide the selector entirely from users who cannot view any branches
+        if not user_has_branch_permission(user):
+            return ''
```
jeremystretch: Can we instead evaluate this inside the HTML template via `perms.netbox_branching.view_branch`?

netbox_branching/views.py
```diff
@@ -633,9 +633,16 @@ def get_required_permission(self):
     def get(self, request):
         return redirect(self.get_return_url(request))
 
+    def _get_form(self, data=None, initial=None):
```
jeremystretch, line 636: Maybe change to a generic signature?
```suggestion
    def _get_form(self, *args, **kwargs):
```

jeremystretch (changes requested):
I like the approach in general, but do think we should

1. Split out the documentation addition to a separate PR, and
2. Ditch `resolve_request_user()`

netbox_branching/utilities.py
arthanson: Kept resolve_request_user() rather than request.user, since the middleware runs before DRF authentication and request.user is still anonymous for token requests, which would 400 every X-NetBox-Branch API call.

netbox_branching/utilities.py
arthanson: Per Claude: The branch decides which schema every query hits, so it must be resolved before the view runs - hence middleware, unlike NetBox's own restrict() calls which all happen inside views. But DRF doesn't authenticate token requests until view dispatch, so at that point request.user is still AnonymousUser. Using request.user directly makes every token request with X-NetBox-Branch return a 400.

netbox_branching/views.py
arthanson: I'd push back here - It's a private helper, so *args, **kwargs only hides the two arguments it actually takes.

arthanson:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Just new changes since last review.

arthanson:
Per Claude review:

1. docs will be split out as separate if needed.
2. A fragility if called differently but not an actionable concern.

PR #662, opened 2026-09-15
#641 - Deprovision the branch before deleting its row

### LINKED ISSUE #641: Branch deletion 500s (taggit ValueError) when an Event Rule subscribes to branch_deprovisioned; deletion rolls back
### Deployment Type
NetBox Cloud

### NetBox Version
v4.6.8

### netbox-branching Version
1.1.3 (bug present in source at `57c7a96` and in `v1.2.0-beta1`)

### Steps to Reproduce
1. Create an enabled Event Rule with object type `netbox_branching.branch` whose event types include `branch_deprovisioned` (e.g. a webhook rule forwarding branch lifecycle events).
2. Create a branch; wait for `ready`.
3. Delete the branch (UI or `DELETE /api/plugins/branching/branches/{id}/`).

### Expected Behavior
Branch row deleted and schema dropped; the `branch_deprovisioned` event rule fires.

### Observed Behavior
HTTP 500: `ValueError: Branch objects need to have a primary key value before you can access their tags.` The deletion rolls back (row delete and schema drop are atomic per #445), so the branch survives in `ready` status. With `MAX_WORKING_BRANCHES`-style caps this makes rejected/abandoned branches effectively immortal: 4/4 deletes failed deterministically on our deployment until the event rule was adjusted.

### Analysis
`Branch.delete()` (`netbox_branching/models/branches.py:359`) runs `super().delete()` first — Django's collector then sets `self.pk = None` — and calls `self.deprovision()` afterwards, inside the same `transaction.atomic()` block. `deprovision()` sends `post_deprovision` (`branches.py:1417`), which `handle_branch_event` receives (`signal_receivers.py:259`); when a matching Event Rule exists it calls `serialize_object(branch)` (`signal_receivers.py:237`) → NetBox `features.py` accesses `self.tags.all()` → django-taggit raises on the pk-less instance. With no matching event rule, `handle_branch_event` returns early (`signal_receivers.py:230`), which is why most installs never see this.

### Suggested fix
Call `self.deprovision()` **before** `super().delete()` inside the same atomic block (atomicity from #445 is preserved — Postgres DDL is transactional), so `post_deprovision` receivers see a branch that still has its pk. Alternatively, serialize the branch before the row delete and pass the snapshot to the event.

Found while operating the NetBox Agents Platform (internal ref AI-1603).

arthanson:
### Fixes: #641 

`Branch.delete()` deleted the row first, nulling the instance pk, then deprovisioned - so the `post_deprovision` receiver failed to serialize the pk-less branch whenever an event rule watched `branch_deprovisioned`, and the 500 rolled back the delete.

Deprovisioning now runs first within the same atomic block. Adds regression tests.


arthanson:

arthanson:
Claude review:

1. Not redundant - serialize_object() returns only Django's fields, which excludes the pk. This line is the payload's sole source of id; removing it breaks all branch events.

2. assertEqual(call_count, 1) already fails at 0, so the no-op case can't occur. Added a failure message instead for clearer diagnostics.

jnovinger (approved):

Issue #445, opened 2026-02-27
Provisioned branch stays in status provisioning due to race condition on edit


MEschenbacher:
### Plugin Version

0.8.1

### NetBox Version

4.4.10

### Python Version

3.13

### Steps to Reproduce

Based on the logs, I believe this is what happened

- A user has creates a branch
- The request queue worker kicks in, provisioning the branch (creating schema, etc.)
- The user now edits the branch, changes the name and clicks *save* exactly around the finalization of the branch provisioning https://github.com/netboxlabs/netbox-branching/blob/v0.8.1/netbox_branching/models/branches.py#L802-L823 this results in a branch that is most likely be provisioned and usable, however still in state provisioning.
- From a user's perspective, the branch is stuck in provisioning. Deletion of this branch is not possible and results in an error along the lines of "deletion of a branch in state provisioning is not possible". The schema, however, will get deprovisioned with the first deletion try in https://github.com/netboxlabs/netbox-branching/blob/v0.8.1/netbox_branching/models/branches.py#L207

The logs

```
# webserver
fd00:2001:0:d00c::4 - ipam.company.net - [25/Feb/2026:23:50:08 +0100] ECDHE-ECDSA-AES256-GCM-SHA384 "GET /plugins/branching/branches/add/ HTTP/1.1" 200 11220 "https://ipam.company.net/plugins/branching/branches/?_branch=" "Mozilla/5.0 (X11; Linux x86_64; rv:147.0) Gecko/20100101 Firefox/147.0" "-"
fd00:2001:0:d00c::4 - ipam.company.net - [25/Feb/2026:23:50:14 +0100] ECDHE-ECDSA-AES256-GCM-SHA384 "POST /plugins/branching/branches/add/ HTTP/1.1" 302 0 "https://ipam.company.net/plugins/branching/branches/add/" "Mozilla/5.0 (X11; Linux x86_64; rv:147.0) Gecko/20100101 Firefox/147.0" "-"
fd00:2001:0:d00c::4 - ipam.company.net - [25/Feb/2026:23:50:15 +0100] ECDHE-ECDSA-AES256-GCM-SHA384 "GET /plugins/branching/branches/7/ HTTP/1.1" 200 11513 "https://ipam.company.net/plugins/branching/branches/add/" "Mozilla/5.0 (X11; Linux x86_64; rv:147.0) Gecko/20100101 Firefox/147.0" "-"
[...]

PR #666, opened 2026-09-16
Closes #665: Add a `provisioned` boolean field on Branch

### LINKED ISSUE #665: Add a `provisioned` boolean field on Branch
### Plugin Version

v1.2.2

### Proposed functionality

Add a BooleanField (default false) on the Branch model named `provisioned`. This non-editable field will automatically be updated by the backend when a branch is provisioned or deprovisioned.

### Use case

Future backends may need to detect whether a branch has already been provisioned (as opposed to only its record having been created in NetBox). Currently, this can be inferred only from the branch's status, which may not accurately indicate when a branch has been deprovisioned successfully.

### External dependencies

N/A

jeremystretch:
Closes #665.

## Summary

Adds a non-editable `provisioned` BooleanField (default `False`) to `Branch`, recording whether a live PostgreSQL schema backing the branch currently exists. It is set when provisioning completes, and cleared when provisioning fails or when the branch is deprovisioned — by archiving or deleting it.

Status cannot answer this reliably: `failed` covers both a provisioning failure, which leaves no schema, and a migration failure, which leaves one fully intact. Accordingly `get_changes()`, `get_unmerged_changes()` and `pending_migrations` now gate on the flag rather than inferring a dataset's existence from status.

## Changes

- **`models/branches.py`** — the field; added to `LIFECYCLE_FIELDS` so an ordinary save can't clobber it (#445); set in `provision()` (both success and failure paths) and cleared in `deprovision()`, guarded by `if self.pk` since `delete()` deprovisions after the row is gone.
- **`migrations/0010_branch_provisioned.py`** — `AddField` plus a backfill that probes `pg_catalog.pg_namespace` for each branch's schema. That is exact where status inference is not. `pg_namespace` rather than `information_schema.schemata`, which lists only schemas the current role owns or holds a privilege on. Marked `fake_on_branch = True`.
- **Exposure** — read-only in `BranchSerializer`, a `?provisioned=` filter, a UI filter dropdown, a non-default table column, and a row on the branch detail page.
- **Docs** — a new field section in `docs/models/branch.md` and the sample response in `docs/rest-api.md`.

## Testing

- 4 new tests covering provision success, provision failure, `deprovision()` and `archive()`; 2 API tests asserting the field is returned and read-only; 1 filterset test.
- Extended `test_edit_form_preserves_lifecycle_fields` (the #445 regression test) to cover the new field.
- `ruff check`, `mkdocs build --strict`, and `makemigrations --check` all clean.

## Notes for review

- **Migration number collision with `618-pluggable-backend`.** That branch already carries `0010_pluggable_backend.py` and its own `0011_branch_provisioned.py`. Whichever lands second needs renumbering, and 618's version of this field will need reconciling rather than merging — its backfill keys on `backend_id` (which doesn't exist here) and it adds a `CheckConstraint` tying `provisioned` to that identifier.
- **Deliberate divergence from 618.** That branch keeps `provisioned` internal (absent from the serializer, listed in the filterset tests' `ignore_fields`). This PR exposes it via the API, filters and UI instead.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jeremystretch:

jeremystretch:
None of the above feedback is actionable. After several rounds of local review prior to the above, I'm comfortable merging this as-is.

PR #667, opened 2026-09-16
#642 -  Return 400 instead of 500 for API requests naming a non-ready branch

### LINKED ISSUE #642: API request with X-NetBox-Branch for a non-ready branch returns 500 AttributeError instead of the intended 400
### Deployment Type
NetBox Cloud

### NetBox Version
v4.6.8

### netbox-branching Version
1.1.3 (present in source at `57c7a96` and in `v1.2.0-beta1`)

### Steps to Reproduce
1. `POST /api/plugins/branching/branches/` — note the returned id/schema_id; the branch is in `new`/`provisioning`.
2. Immediately issue any REST API request carrying `X-NetBox-Branch: <schema_id>` (e.g. `POST /api/ipam/prefixes/{id}/available-ips/`), before the branch reaches `ready`.

### Expected Behavior
HTTP 400 with the existing message: `Branch <name> is not ready for use (status: provisioning)`.

### Observed Behavior
HTTP 500: `AttributeError: 'HttpResponseBadRequest' object has no attribute 'schema_name'`. The intended 400 message is swallowed, so API clients (and tooling built on them) see an inscrutable server error instead of "retry when the branch is ready".

### Analysis
`get_active_branch()` (`netbox_branching/utilities.py:527-535`) **returns** an `HttpResponseBadRequest(...)` object on the API-header path when the branch is not ready, instead of raising. Two consumers then mishandle that value:
- `BranchMiddleware.__call__` (`middleware.py:54-58`) only catches `ObjectDoesNotExist` and assigns the response object to `request.active_branch` with no isinstance check;
- `ActiveBranchContextManager` (`utilities.py:611-616`) does `if (branch := get_active_branch(request)): return activate_branch(branch)` — the response object is truthy, and `activate_branch()` reads `branch.schema_name` → AttributeError → 500.

### Suggested fix
Have `get_active_branch()` raise a dedicated exception for "branch not ready" and map it to a 400 in `BranchMiddleware` (mirroring the existing `ObjectDoesNotExist` → `HttpResponseBadRequest("Invalid branch identifier")` handling); or isinstance-check the return value in both consumers.

Found while operating the NetBox Agents Platform (internal ref AI-1605).

arthanson:
### Fixes: #642 

`get_active_branch()` returned an `HttpResponseBadRequest` object instead of a Branch, and the request processor activated it as one. The first branch-aware query then failed with `'HttpResponseBadRequest' object has no attribute 'schema_name'` (HTTP 500), while endpoints issuing no branchable query silently served main.

It now records the non-ready branch on the request and returns `None`, so nothing is activated, and `BranchMiddleware` rejects the request with the intended 400. The message carries the branch's status but not its name, since the response precedes REST API authentication.


arthanson:

arthanson:
Per claude review: fixed 1 and 2

jeremystretch:
Closing this in favor of PR #673

PR #669, opened 2026-09-17
feat: declare release metadata in netbox-plugin.yaml

joelstoddard:
## What
- Add `netbox-plugin.yaml` (schema v1) at the repo root, declaring `package_name`, `plugin_name`, `netbox_min`/`netbox_max`, and the plugin's `config` (`required_settings`, `default_settings`).
- Add `MANIFEST.in` so `netbox-plugin.yaml` ships inside the sdist.

## Why
NetBox Labs runs a plugin catalog that reads release metadata about each plugin. Until now it has had three channels to build that metadata from: a fallback based on a git tag, inference from the published package, and this "declared" channel — data the plugin's own maintainers assert directly. The declared channel is new and currently has no producers. This PR gives it its first one for `netbox_branching`, and establishes the pattern that will be applied to the other NetBox Labs plugins under CAP-171.

The declared channel outranks what the catalog would otherwise infer from the package, so getting these values right matters. Merging this PR changes no runtime behavior in the plugin itself, and nothing reaches the catalog until the next release is published — the catalog only reads `netbox-plugin.yaml` from a built release artifact.

## How
Every value in `netbox-plugin.yaml` was read directly from this repo, not assumed:
- `plugin_name` (`netbox_branching`), `netbox_min` (`4.7.0`) and `netbox_max` (`4.7.99`) come from `netbox_branching/__init__.py` (`AppConfig.name`, `min_version`, `max_version`).
- The 16 `default_settings` entries are copied from the same file's `AppConfig.default_settings` dict, values and defaults unchanged.
- `required_settings: []` is written explicitly (the repo has no `required_settings` attribute anywhere), since the `config` block asserts that field either way and being explicit documents the intent.
- `package_name` is `netboxlabs-netbox-branching`, the published PyPI name.
- `requires_plugins` and `conflicts` are omitted rather than written as `[]` — there's no evidenced inter-plugin dependency, and omitting signals "nothing to add" rather than clearing a future inferred/validated entry.

Validated with `netbox-plugin-lint netbox-plugin.yaml --strict` (clean, no warnings — this plugin has no loopback-style default value to trip the strict check), and confirmed `netbox-plugin.yaml` is present in the built sdist (`uv build --sdist` + `tar -tzf`) before committing, so `MANIFEST.in` is proven to work rather than assumed.

Not run: the repository's own test suite (`python netbox/manage.py test netbox_branching.tests`) — it requires a live NetBox install and will not run cleanly against a bare clone, so it was not attempted here. `ruff check` was run and surfaced 5 pre-existing findings, all in files this PR does not touch.

If any of the declared values above are wrong or the `netbox_min`/`netbox_max` bounds don't reflect this plugin's actual compatibility policy, please correct them directly — this was written by someone outside the team that owns this repo, working from the source rather than institutional knowledge.


joelstoddard:
Closing: opened from a personal fork by mistake. CAP-171 will re-land this on a branch once write access to this repo is sorted. The declared content is unchanged and validated.

Issue #670, opened 2026-09-18
Resolve foreign-key ids in Change Diff view


gzahl:
### Plugin Version

1.2.1

### Proposed functionality

The Change Diff view only shows numeric foreign key ids for fields like "cluster", "platform" etc in the "Branch (current)", "Branch (main)".
It would be much easier to review a branch if we only or additionally would get the name of the object instead of their numeric id.

### Use case

We use the Change Diff to review a branch before merging. The numeric ids can not be easily be reviewed by a human.

PR #671, opened 2026-09-18
Resolve fk ids changediff #670

mjung-shnetz:
Resolves #670 by showing the string representation of related objects (e.g. `platform`, `cluster`) instead of raw numeric foreign-key IDs in the Change Diff detail view.

## Changes
- Added `ChangeDiff.original_display` / `modified_display` / `current_display` cached properties, which resolve ForeignKey/OneToOneField/ManyToManyField values against the correct schema (branch schema for `modified`, main for `original`/`current`).

- Updated `ChangeDiffView` to build the branch/main diff summaries from the resolved values.

- Updated the `changediff.html` template (data columns and conflicts table) to render the resolved values.

- Added test coverage in `test_changediff.py`, including a branch-schema-specific case for objects that only exist in the branch.


jnovinger:
@mjung-shnetz, thank you for your interest in contributing to netbox-branching, however, it appears there is no accepted issue that correlates to this pull request. Before pull requests are opened, we require an accepted issue. Please wait for your FR to be reviewed and accepted before further work is done on this pull request.

I'm going to close this PR for the moment, but we can re-open it if we decide it makes sense to adopt the FR.

Issue #672, opened 2026-09-18
`get_active_branch()` returns an `HttpResponseBadRequest` instead of a Branch


jeremystretch:
### Plugin Version

1.2.1

### NetBox Version

4.7.1

### Python Version

3.14

### Steps to Reproduce

1. Create a branch named `Branch 1` and allow it to finish provisioning (status `ready`).
2. Put the branch into a transitional status, as it would be during a background operation. Either start a sync and issue the next step while it runs, or set the status directly:
   ```
   python netbox/manage.py nbshell
   >>> from netbox_branching.models import Branch
   >>> from netbox_branching.choices import BranchStatusChoices
   >>> Branch.objects.filter(name='Branch 1').update(status=BranchStatusChoices.SYNCING)
   ```
3. Issue a REST API request that activates the branch by header, against any branch-aware endpoint:
   ```
   curl -H "Authorization: Token $TOKEN" \
        -H "X-NetBox-Branch: $SCHEMA_ID" \
        http://netbox:8000/api/dcim/sites/
   ```

The same applies to GraphQL requests, which `is_api_request()` also covers.

### Expected Behavior

HTTP 400, with the body `Branch Branch 1 is not ready for use (status: syncing)` — which is the response `get_active_branch()` constructs for exactly this case.

### Observed Behavior

HTTP 500. An `AttributeError` is raised out of the database router:

```
  File "netbox_branching/database.py", line 47, in db_for_read
    return self._get_db(model, **hints)
  File "netbox_branching/database.py", line 36, in _get_db
    return self._get_connection(branch)
  File "netbox_branching/database.py", line 22, in _get_connection
    return f'{self.connection_prefix}{branch.schema_name}'
                                      ^^^^^^^^^^^^^^^^^^
AttributeError: 'HttpResponseBadRequest' object has no attribute 'schema_name'
```

The intended 400 is never returned; the failure surfaces far from its cause, in the router, on the first query against a branchable model.

Only the API-header path is affected. The UI query-parameter path (`?_branch=<schema_id>`) handles an unready branch correctly — it adds an error message and returns `None`.

### Suspected Cause

`get_active_branch()` in `netbox_branching/utilities.py` returns a *response object* on one of its paths:

```python
if is_api_request(request) and BRANCH_HEADER in request.headers:
    branch = Branch.objects.get(schema_id=request.headers.get(BRANCH_HEADER))
    if not branch.ready:
[...]

PR #673, opened 2026-09-18
Fixes #672: Refuse an unready branch instead of activating a 400 response

### LINKED ISSUE #642: API request with X-NetBox-Branch for a non-ready branch returns 500 AttributeError instead of the intended 400

jeremystretch:
### Fixes: #642

Note: This ended up being a duplicate to PR #667, but offers a more comprehensive fix.

## Summary

`get_active_branch()` returned an `HttpResponseBadRequest` on the API-header path when the named branch existed but was not ready. Both of its callers treat the return value as a `Branch`, and an `HttpResponse` is truthy, so the response object was installed as the active branch rather than returned to the client:

- `ActiveBranchContextManager()` put it in the `active_branch` context variable, and
- `BranchMiddleware.__call__()` assigned it to `request.active_branch`.

The first query against a branchable model then read `branch.schema_name` off it in `BranchAwareRouter._get_connection()`:

```
AttributeError: 'HttpResponseBadRequest' object has no attribute 'schema_name'
```

So the intended 400 surfaced as a 500, from the router, far from its cause.

## Changes

- **`utilities.py`** — new `BranchNotReady` exception, raised in place of returning a response. `get_active_branch()` now has a single contract: a `Branch`, or `None`, or raise.
- **`middleware.py`** — catches `BranchNotReady` and returns the 400 the old code meant to return, preserving today's status code and message.
- **`utilities.py`** — `ActiveBranchContextManager()` catches it too and declines to activate.

That last one is not redundant. The context manager is registered as a NetBox request processor, applied by `CoreMiddleware`, which sits ahead of plugin middleware in `MIDDLEWARE` — so it reaches `get_active_branch()` before `BranchMiddleware` does and cannot rely on the middleware to keep an unusable branch out of the context variable. Catching it explicitly also avoids `apply_request_processors()` reporting an expected refusal as a failed request processor (its blanket `except Exception` would otherwise `warnings.warn` on every such request).

Only the API-header path was affected. The query-parameter path already returned `None` for an unready branch, and is unchanged.

## Tests

Two regression tests in `tests/test_request.py`, ported from the equivalent fix on the `v2.0` line (#654):

- `test_api_header_with_unready_branch_returns_400` — the end-to-end contract.
- `test_unready_branch_is_never_activated` — pins the request-processor half, which the middleware test alone would not cover.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jeremystretch:

arthanson (approved):

PR #674, opened 2026-09-18
#469 - Retrace cable paths after a squash merge

### LINKED ISSUE #469: Squash merge does not recalculate cable paths

arthanson:
### Fixes: #469 

Squash merge applies a create with a raw save, which bypasses `Model.save()` and the dependent objects it maintains — for a Cable, its `CablePath` records. Iterative merge gets them incidentally, from the trailing UPDATE change it replays through a real `save()`; squash collapses that away.

After every change has been applied, the merge now calls `update_dependent_objects()` on each created object whose model defines it (`hasattr`-guarded, as with `deserialize_object()`). The same pass runs after a revert, for objects a restored delete brought back. Nothing model-specific lives in the plugin.

Un-skips `test_merge_cable_path_recalculation`, and adds `test_revert_cable_path_recalculation` for the revert case.

Requires netbox-community/netbox#23209, which adds the hook and implements it on `Cable`. Until that merges, CI needs the `test-against:21879-update-dependent` label; without the hook the guard skips and both cable tests fail.


arthanson:

arthanson:
Per Claude review - the bug and fix are only relevant to squash merge.

